PR #9, opened 2022-03-18
Tighter dependencies with features

djeedai:
Enable minimal dependencies with new features `bevu_sprite` and
`bevy_ui`, removing the `bevy/render` mandatory dependency. Those new
features are enabled by default, and enable the built-in lenses for the
related Bevy crates. The core `bevy_tweening` crate itself does not take
any optional Bevy dependency anymore, allowing for a slim build with
only the core Bevy functionalities.


djeedai:
CI hitting https://github.com/rust-lang/rust/issues/94998, witing 24h for new rustc nightly to retry.

PR #11, opened 2022-04-23
Updated all document links to point to bevy 0.7.

iancormac84:


djeedai:
Thanks @iancormac84 for the fix! Could you please rebase on latest; I fixed the example that was making CI fail, so it should all go green for you too now. Thanks!

djeedai (approved):

PR #12, opened 2022-04-23
Only test docs with all features

djeedai:
Some examples in docs depend on `Sprite`, only available with the
`bevy_sprite` feature, so fail CI. Ignore doc testing in all variants
except the `--all-features` one.


djeedai:
Abandoning in favor of #13 which is better for local build, as it supports `--no-default-features`.

PR #15, opened 2022-04-24
Add `Tween::set_direction()` for backward playback

djeedai:
Allow a `Tween` to play backward by setting its direction with
`Tween::set_direction()`.


djeedai:
This is just a subset of #17, for the particular case of `Tween` itself. This needs to be extended to other tweenables for #17 to be completed.

Issue #8, opened 2022-03-06
Extend Tracks and Sequence functionality


Shatur:
To combine multiply tweens and loop them I have to create custom lens which is not very ergonomic. We currently have `Tracks` and `Sequence`, but they aren't loopable.
I would like to request to extend their functionality. It would be great if we could combine multiply tweens with different tweening type. For example, to animate healing pickup like in Owerwatch I need rotation tween with `Loop` and translation tween with `PingPong`. [Video](https://www.youtube.com/watch?v=bT-VwsOIvR4) of how such pickups look.

djeedai:
Thanks for logging this and for the video. It's definitely something I want to add.

djeedai:
Note: I've looked into looping `Tracks` and it's a minefield of corner cases that I'm not sure how to handle. For example with one tweenable of 0.8 second and another one of 1 second, what does looping look like in the [0.8:1.0] time frame for the first tweenable (since the `Tracks` will be 1 second long, the max of all its children)? What about ping-pong looping (reverse playback) in that time frame? etc.

Shatur:
@djeedai Makes sense... Maybe we should keep only `Tween` objects loopable, but turn `Tracks` into a simple array to allow users to apply multiply `Tween`s to the same object?

djeedai:
@Shatur I'm not sure I follow, that sounds like what `Tracks` is already doing. What is the difference between what you describe and what `Tracks` currently does, which is applying a collection of `Tweenable` in parallel to the same object. Granted, it doesn't have introspection so you can't index like an array to inspect its content, but internally it is an array.

Shatur:
@djeedai maybe I get it wrong, but If I understand correctly, I can't execute multiply loopable tweens in parallel with Tracks. Is it correct? I would expect `Tracks` to have infinity duration if at least one `Tween` is loopable.

djeedai:
> If I understand correctly, I can't execute multiply loopable tweens in parallel with `Tracks`. Is it correct?

Correct.

> I would expect `Tracks` to have infinity duration if at least one Tween is loopable.

Yes. But that's the easy part of making `Tracks` loopable. Are you saying you want each single tweenable to loop completely independently of each other, and so `Tracks` wouldn't really have a defined duration anymore? So `Tracks` wouldn't implement `Tweenable`.

Shatur:
> So Tracks wouldn't implement Tweenable.

[...]

PR #19, opened 2022-05-07
Support running an animation N times

### LINKED ISSUE #17: Allow tweenables to play backward
Allow all `Tweenable` to play backward once or looped, or to start from end in ping-pong mode.

Also allow this direction to be changed at runtime at any point, to allow some animation to "go back to start" without having to create a new `Tweenable` which encodes the reverse animation explicitly.

Note: this is a prerequisite for #16

SUPERCILEX:
Fixes #17

This supports an animation like the following:

```rust
pub fn error_shake(current: &Transform) -> Sequence<Transform> {
    let wiggle = Quat::from_rotation_z(PI / 16.);
    Tween::new(
        EaseMethod::Linear,
        TweeningType::Once,
        Duration::from_millis(25),
        TransformRotationLens {
            start: current.rotation,
            end: current.rotation * wiggle.inverse(),
        },
    )
    .then(Tween::new(
        EaseMethod::Linear,
        TweeningType::PingPongTimes(3),
        Duration::from_millis(50),
        TransformRotationLens {
            start: current.rotation * wiggle.inverse(),
            end: current.rotation * wiggle,
        },
    ))
    .then(Tween::new(
        EaseMethod::Linear,
        TweeningType::Once,
        Duration::from_millis(25),
        TransformRotationLens {
            start: current.rotation * wiggle,
            end: current.rotation,
        },
    ))
}
```

### DIFF
--- a/examples/sequence.rs
+++ b/examples/sequence.rs
@@ -1,6 +1,8 @@
-use bevy::prelude::*;
-use bevy_tweening::{lens::*, *};
 use std::time::Duration;
+
+use bevy::prelude::*;
+
+use bevy_tweening::{lens::*, *};
 
 fn main() {
     App::default()
@@ -107,15 +109,16 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
         Vec3::new(margin, screen_y - margin, 0.),
         Vec3::new(margin, margin, 0.),
     ];
-    // Build a sequence from an iterator over a Tweenable (here, a Tween<Transform>)
+    // Build a sequence from an iterator over a Tweenable (here, a
+    // Tracks<Transform>)
     let seq = Sequence::new(dests.windows(2).enumerate().map(|(index, pair)| {
         Tracks::new([
             Tween::new(
                 EaseFunction::QuadraticInOut,
                 Duration::from_millis(250),
-                TransformRotationLens {
-                    start: Quat::IDENTITY,
-                    end: Quat::from_rotation_z(180_f32.to_radians()),
+                TransformRotateZLens {
+                    start: 0.,
+                    end: 180_f32.to_radians(),
                 },
             )
             .with_repeat_count(RepeatCount::Finite(4))
--- a/examples/sequence.rs
+++ b/examples/sequence.rs
@@ -119,7 +119,7 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
                 },
             )
             .with_repeat_count(RepeatCount::Finite(4))
-            .with_repeat_strategy(RepeatStrategy::Bounce),
+            .with_repeat_strategy(RepeatStrategy::MirroredRepeat),
             Tween::new(
                 EaseFunction::QuadraticInOut,
                 Duration::from_secs(1),
--- a/examples/sequence.rs
+++ b/examples/sequence.rs
@@ -111,16 +111,27 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
     ];
     // Build a sequence from an iterator over a Tweenable (here, a Tween<Transform>)
     let seq = Sequence::new(dests.windows(2).enumerate().map(|(index, pair)| {
-        Tween::new(
-            EaseFunction::QuadraticInOut,
-            TweeningType::Once,
-            Duration::from_secs(1),
-            TransformPositionLens {
-                start: pair[0] - center,
-                end: pair[1] - center,
-            },
-        )
-        .with_completed_event(true, index as u64) // Get an event after each segment
+        Tracks::new([
+            Tween::new(
+                EaseFunction::QuadraticInOut,
+                Duration::from_millis(250),
+                TransformRotationLens {
+                    start: Quat::IDENTITY,
+                    end: Quat::from_rotation_z(180_f32.to_radians()),
+                },
+            )
+            .with_repeat_count(RepeatCount::Finite(4))
+            .with_repeat_strategy(RepeatStrategy::Bounce),
+            Tween::new(
+                EaseFunction::QuadraticInOut,
+                Duration::from_secs(1),
+                TransformPositionLens {
+                    start: pair[0] - center,
+                    end: pair[1] - center,
+                },
+            )
+            .with_completed_event(true, index as u64), // Get an event after each segment
+        ])
     }));
 
     commands
@@ -139,7 +150,6 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
     // size at the same time.
     let tween_move = Tween::new(
         EaseFunction::QuadraticInOut,
-        TweeningType::Once,
         Duration::from_secs(1),
         TransformPositionLens {
             start: Vec3::new(-200., 100., 0.),
@@ -149,7 +159,6 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
     .with_completed_event(true, 99); // Get an event once move completed
     let tween_rotate = Tween::new(
         EaseFunction::QuadraticInOut,
-        TweeningType::Once,
         Duration::from_secs(1),
         TransformRotationLens {
             start: Quat::IDENTITY,
@@ -158,7 +167,6 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
     );
     let tween_scale = Tween::new(
         EaseFunction::QuadraticInOut,
-        TweeningType::Once,
         Duration::from_secs(1),
         TransformScaleLens {
             start: Vec3::ONE,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,10 +42,7 @@
 //! let tween = Tween::new(
 //!     // Use a quadratic easing on both endpoints.
 //!     EaseFunction::QuadraticInOut,
-//!     // Loop animation back and forth.
-//!     TweeningType::PingPong,
-//!     // Animation time (one way only; for ping-pong it takes 2 seconds
-//!     // to come back to start).
+//!     // Animation time.
 //!     Duration::from_secs(1),
 //!     // The lens gives access to the Transform component of the Entity,
 //!     // for the Animator to animate it. It also contains the start and
@@ -85,7 +82,6 @@
 //! let tween1 = Tween::new(
 //!     // [...]
 //! #    EaseFunction::BounceOut,
-//! #    TweeningType::Once,
 //! #    Duration::from_secs(2),
 //! #    TransformScaleLens {
 //! #        start: Vec3::ZERO,
@@ -95,7 +91,6 @@
 //! let tween2 = Tween::new(
 //!     // [...]
 //! #    EaseFunction::QuadraticInOut,
-//! #    TweeningType::Once,
 //! #    Duration::from_secs(1),
 //! #    TransformPositionLens {
 //! #        start: Vec3::ZERO,
@@ -163,22 +158,34 @@ pub use plugin::{
 };
 pub use tweenable::{Delay, Sequence, Tracks, Tween, TweenCompleted, TweenState, Tweenable};
 
-/// Type of looping for a tween animation.
+/// How many times to repeat a tween animation. See also: [`RepeatMode`].
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum TweeningType {
-    /// Run the animation once from start to end only.
-    Once,
-    /// Loop the animation indefinitely, restarting from the start each time the end is reached.
-    Loop,
-    /// Loop the animation back and forth, changing direction each time an endpoint is reached.
-    /// A complete cycle start -> end -> start always counts as 2 loop iterations for the various
-    /// operations where looping matters.
-    PingPong,
+pub enum RepeatCount {
+    /// Run the animation N times.
+    Finite(u32),
+    /// Loop the animation indefinitely.
+    Infinite,
 }
 
-impl Default for TweeningType {
+/// What to do when a tween animation needs to be repeated.
+///
+/// Only applicable when [`RepeatCount`] is greater than 1.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum RepeatStrategy {
+    /// Reset the animation back to its starting position.
+    #[default]
+    Teleport,
+    /// Follow a ping-pong pattern, changing the direction each time an endpoint is reached.
+    ///
+    /// A complete cycle start -> end -> start always counts as 2 loop iterations for the various
+    /// operations where looping matters. That is, a 1 second animation will take 2 seconds to end
+    /// up back where it started.
+    Bounce,
+}
+
+impl Default for RepeatCount {
     fn default() -> Self {
-        TweeningType::Once
+        Self::Finite(1)
     }
 }
 
@@ -634,9 +641,15 @@ mod tests {
     }
 
     #[test]
-    fn tweening_type() {
-        let tweening_type = TweeningType::default();
-        assert_eq!(tweening_type, TweeningType::Once);
+    fn repeat_count() {
+        let tweening_type = RepeatCount::default();
+        assert_eq!(tweening_type, RepeatCount::Finite(1));
+    }
+
+    #[test]
+    fn repeat_strategy() {
+        let tweening_type = RepeatStrategy::default();
+        assert_eq!(tweening_type, RepeatStrategy::Teleport);
     }
 
     #[test]
@@ -686,7 +699,6 @@ mod tests {
     fn animator_new() {
         let tween = Tween::new(
             EaseFunction::QuadraticInOut,
-            TweeningType::PingPong,
             std::time::Duration::from_secs(1),
             DummyLens { start: 0., end: 1. },
         );
@@ -702,7 +714,6 @@ mod tests {
         for state in [AnimatorState::Playing, AnimatorState::Paused] {
             let tween = Tween::<DummyComponent>::new(
                 EaseFunction::QuadraticInOut,
-                TweeningType::PingPong,
                 std::time::Duration::from_secs(1),
                 DummyLens { start: 0., end: 1. },
             );
@@ -720,7 +731,6 @@ mod tests {
 
         let tween = Tween::<DummyComponent>::new(
             EaseFunction::QuadraticInOut,
-            TweeningType::PingPong,
             std::time::Duration::from_secs(1),
             DummyLens { start: 0., end: 1. },
         );
@@ -734,7 +744,6 @@ mod tests {
     fn animator_controls() {
         let tween = Tween::<DummyComponent>::new(
             EaseFunction::QuadraticInOut,
-            TweeningType::PingPong,
             std::time::Duration::from_secs(1),
             DummyLens { start: 0., end: 1. },
         );
@@ -773,7 +782,6 @@ mod tests {
     fn asset_animator_new() {
         let tween = Tween::<DummyAsset>::new(
             EaseFunction::QuadraticInOut,
-            TweeningType::PingPong,
             std::time::Duration::from_secs(1),
             DummyLens { start: 0., end: 1. },
         );
@@ -789,7 +797,6 @@ mod tests {
         for state in [AnimatorState::Playing, AnimatorState::Paused] {
             let tween = Tween::<DummyAsset>::new(
                 EaseFunction::QuadraticInOut,
-                TweeningType::PingPong,
                 std::time::Duration::from_secs(1),
                 DummyLens { start: 0., end: 1. },
             );
@@ -809,7 +816,6 @@ mod tests {
 
         let tween = Tween::new(
             EaseFunction::QuadraticInOut,
-            TweeningType::PingPong,
             std::time::Duration::from_secs(1),
             DummyLens { start: 0., end: 1. },
         );
@@ -824,7 +830,6 @@ mod tests {
     fn asset_animator_controls() {
         let tween = Tween::new(
             EaseFunction::QuadraticInOut,
-            TweeningType::PingPong,
             std::time::Duration::from_secs(1),
             DummyLens { start: 0., end: 1. },
         );
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -170,10 +170,9 @@ pub enum RepeatCount {
 /// What to do when a tween animation needs to be repeated.
 ///
 /// Only applicable when [`RepeatCount`] is greater than 1.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum RepeatStrategy {
     /// Reset the animation back to its starting position.
-    #[default]
     Teleport,
     /// Follow a ping-pong pattern, changing the direction each time an endpoint is reached.
     ///
@@ -186,6 +185,12 @@ pub enum RepeatStrategy {
 impl Default for RepeatCount {
     fn default() -> Self {
         Self::Finite(1)
+    }
+}
+
+impl Default for RepeatStrategy {
+    fn default() -> Self {
+        Self::Teleport
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -175,14 +175,14 @@ pub enum RepeatCount {
     /// Run the animation N times.
     Finite(u32),
     /// Run the animation for some amount of time.
-    Until(Duration),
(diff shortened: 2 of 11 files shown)


src/lib.rs
```diff
@@ -167,10 +167,14 @@ pub enum TweeningType {
     Once,
     /// Loop the animation indefinitely, restarting from the start each time the end is reached.
     Loop,
+    /// [Loop][TweeningType::Loop]s N times.
+    LoopTimes(u32),
```
djeedai, line 171: I've come up with the same conclusion that we need something like that while working on making `Sequence` loopable, but I think there's a better design than that. Adding even more tweening types just makes handling them all more complex.

Instead, we can decouple the concepts of _repeat count_ on one side, and _repeat mode_ on the other side.
- The repeat count is self-explanatory.
- The repeat mode is simply `Normal` or `Mirrored`, describes how the playback direction changes (or not) when reaching the end of the current playback cycle/loop.

All existing `TweeningType` enum values, as well as the ones added in this PR, can be described in terms of those two quantities. And I'm pretty sure this makes some operations easier to write internally, as we sometimes care about the repeat count or the mirroring, but rarely both at once.

Note that although I'm grateful for the contribution, and always open to them, it's probably better to start a discussion on Discord or open a GitHub issue to discuss this kind of change before jumping into an implementation. Same for the speed feature one.

SUPERCILEX: Oh interesting, so does that mean PingPong can be deleted since you could combine Loop with Mirrored?

I'll let the speed PR settle before working on the others. As for why I went straight to PRs, I needed all this code for my game and am already using it. :grin:

djeedai: Yes indeed, `Once` is "Repeat forward 1 time" (`repeat=Times(1), mode=Normal`), `Loop` is "Repeat forward infinite times" (`repeat=Infinite, mode=Normal`), and `PingPong` is "Repeat mirrored infinite times" (`repeat=Infinite, mode=Mirrored`).

SUPERCILEX: That's super smart, I love it :heart_eyes:

src/tweenable.rs
```diff
@@ -359,6 +352,29 @@ impl<T> Tween<T> {
         self
     }
 
+    /// TODO
+    pub fn set_repeat_count(&mut self, count: RepeatCount) {
+        self.count = count;
+        self.clock.is_looping = count != RepeatCount::Finite(1);
+    }
```
SUPERCILEX, line 359: I like these methods from an initialization devx standpoint, but I have no idea what it means to change the repeat values in the middle of a running animation. Not sure what to do about that.

src/tweenable.rs
```diff
@@ -404,7 +420,11 @@ impl<T> Tweenable<T> for Tween<T> {
     }
 
     fn is_looping(&self) -> bool {
-        self.tweening_type != TweeningType::Once
+        match self.count {
+            RepeatCount::Finite(times) if times == 1 => false,
+            RepeatCount::Finite(times) => self.times_completed < times,
+            RepeatCount::Infinite => true,
+        }
```
SUPERCILEX, line 427: I kept this from the original PR, but I'm not sure if it's the best approach. Currently, there's no way to seek to a specific number of completions: the AnimClock always treats progress as within the 0-1 range of a single run. We could use the truncated part to represent the number of completions, but there are only 23 mantissa bits and we support 32 bits of completion so that doesn't really work.

I'm currently think all the `is_looping` stuff should be removed and an internal `AnimCompletion` class should be created to track when the overall animation is completed (while keeping `AnimClock` in charge of an individual run's completions). Later if there's demand, an API could be added to `Tweenable` that lets you set the number of completions. WDYT?

SUPERCILEX: Note that this would mean set_progress always uses `fract` instead of clamping.

src/tweenable.rs
SUPERCILEX: Thinking about this a bit more, it's probably fine with the is_looping finikyness removed. It just means you can extend or shorten the animation.

src/tweenable.rs
djeedai: > We could use the truncated part to represent the number of completions, but there are only 23 mantissa bits and we support 32 bits of completion so that doesn't really work.

I don't think that's a problem. I have a hard time imagining anyone requesting an animation to repeat exactly 2^26 times, not one more or less. But this raises another issue, which is that the larger the number the less precision in the fractional part (that's how floats work) and so `set_progress()` as is, taking into account the integral part, is not ideal.

Instead I think we should move to `set_progress()` being in [0:1] always (or even [0:1[ if looping), and provide another way to set the times completed. In particular, it might make sense to directly allow setting the elapsed time, which is encoded as `u128` (via `Duration`) so does not suffer from variable precision. Thinking way ahead of time, that would also allow integrating with a timeline UI in some Editor.

> I'm currently think all the `is_looping` stuff should be removed and an internal `AnimCompletion` class should be created to track when the overall animation is completed (while keeping `AnimClock` in charge of an individual run's completions).

I'm not sure I see what that looks like. The point of `AnimClock` is that it should make it easier to handle the case where `tick()` crosses one (or, less likely, several) loop boundary/ies. I'm not sure what `AnimCompletion` would provide beyond what `times_completed : u32` already gives us.

For the case of `is_looping()`, I think it makes few sense from the user's perspective; what makes more sense is querying the repeat count. And internally although we generally still need to know if the tweenable needs to wrap around at the end of this loop, we're not interested in whether the animation is looping, but rather whether the current loop is the last one (completed) or if there's another one after that (wrap around, still active). So either way `is_looping()` probably doesn't make sense anymore once we support a repeat count.

SUPERCILEX: Ok, I got rid of is_looping and moved towards clamp only support. For now I haven't added an API to set the completion count since I'd rather wait until a use case shows up which we can design against.

I also added AnimCompletion. The idea is just to have something keep track of when the animation is finished so that we can use it for sequences and tracks too.

SUPERCILEX:
This PR depends on https://github.com/djeedai/bevy_tweening/pull/29 so that I don't have to manually adjust line breaks while editing the docs.

SUPERCILEX:
Mmmm, the diff is garbage so I guess #29 needs to go through and then I'll rebase.

SUPERCILEX:
Ok, the diff is actually usable now

SUPERCILEX:
@djeedai bump

src/tweenable.rs
```diff
@@ -1338,6 +1366,7 @@ mod tests {
         assert_eq!(tracks.times_completed(), 0);
         assert!(tracks.progress().abs() < 1e-5);
 
+        tracks.rewind();
```
djeedai, line 1369: Why is that needed? It shouldn't, `set_progress()` should take care of updating any state as needed such that the `Tracks` tweenable is in a state where the next `tick()` will start from 90% progress.

djeedai (changes requested):
Thanks for the work! Sorry I missed your earlier message, I didn't realize this was ready for review. I think it looks mostly good, aside from a terminology question on the repeat strategy. However I'm concerned about the additional `reset()` in the examples, it really shouldn't be needed.

src/lib.rs
```diff
@@ -174,24 +165,45 @@ pub use tweenable::{
...
-    /// end is reached.
-    Loop,
-    /// Loop the animation back and forth, changing direction each time an
-    /// endpoint is reached. A complete cycle start -> end -> start always
-    /// counts as 2 loop iterations for the various operations where looping
-    /// matters.
-    PingPong,
+pub enum RepeatStrategy {
+    /// Reset the animation back to its starting position.
+    Teleport,
```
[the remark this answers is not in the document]
SUPERCILEX: Repeat and mirrored repeat works for me.

src/tweenable.rs
SUPERCILEX: Maybe I misunderstood our previous discussion, but I thought the plan was to make progress a local concept. That is, progress only applies within one repetition. The rewinds are there because setting the progress isn't changing the number of times the animation has repeated. Right now, there's no API to do that.

djeedai: The problem of doing so is to implement `Tracks`. I originally added `set_progress()` mainly for that purpose, and exposed it publicly because why not.

Consider a `Track` with 2 `Tween`s:
- `Tween` A repeats every 1 second
- `Tween` B lasts 4 seconds

The `Track` duration is computed as 4 seconds. Now, if you call `set_progress(0.75)` on the track itself, you need somehow to set the `times_completed` of A to 3. The idea was that A's own `set_progress()` internal call should take care of that.

Now, we can also add a separate API to set the `times_completed`, and maybe that's valuable for the case you just want to add a few loops without changing the current state (why, I'm not sure, but let's assume there's a use case). But now for the use case of "I have a duration obtained from my game/editor and want to set the animation progress" it would much nicer if I don't have to do some modulo arithmetic as a user and can just call `set_progress(23.58)` and it sets `times_completed` to 23 and `progress` to 0.58 by itself, no?

SUPERCILEX: Interesting, I only considered sequences where the model works fine since the progress applies within one repetition of the sequence.

I'm not sure your model works because what if Tween A repeats infinitely? The tracks still need to handle setting times completed manually to be correct (since progress * infinity would be infinite repetitions). The model breaks down when you have an infinite tween, hence keeping progress local only.

Then again, I see your point about making it easy to scrub through a tween, but I still really don't like the idea of having to share your internal progress precision with your times completed. I want to get the bits right, so that leaves us with several things to consider if we go the progress does everything route:
- Currently times_completed is 32 bits and everything we do is saturating.
  - Everything should maybe wrap instead? The transition might be glitchy, but at least the animation won't get stuck in infinite mode. A problem with this is that if your repetition count is right on the border of wrapping, the timing might work out such that the tick is too big and causes two completions thus wrapping and causing us to not realize we were supposed to stop. I think this is acceptable given enough bits because no one is going to be repeating exactly some number of times greater than 1000.
  - We can either make times_completed 16 bits and keep f32s which gives 7 decimal places to work with at saturation or use f64 with 32 bits of time completion which gives 11 decimal places of precision at saturation.
- If the way to set times_completed is through progress, then the `times_completed` API needs to be removed.

What do you think about this plan: use 16 bits of times_completed, keep f32 progress, make progress always use fract instead of clamping (aka be times_completed.progress), get rid of the times_completed API.

PS: In terms of PRs, this one is getting too big for me so I'd like to get agreement on future plans, merge it, and then fix whatever the plan is + sequences + tracks in separate PRs which means things might be in a slightly broken state in the interim.

SUPERCILEX: BTW, I think this solves #8. The answer to _what happens with PingPong asymmetric tracks_ is "compute the times_completed.progress for each track given that 1.0 is defined as the length of the longest track." This works fine with infinitely repeating tweens too since we only care about how long a single repetition is.

djeedai: > I only considered sequences where the model works fine since the progress applies within one repetition of the sequence.

Yes `Sequence` is generally easier, `Tracks` is where most issues occur.

> I'm not sure your model works because what if Tween A repeats infinitely?

This is explicitly disallowed. I'm pretty sure I documented that somewhere. That's one limitation of `Tracks`.

> I still really don't like the idea of having to share your internal progress precision with your times completed.

That's a fair point. Setting progress as `f32` will have a limited precision. Although I doubt in practice this matters too much. Note however that I didn't say we should internally merge the [0:1] progress with the `times_completed`; that would be indeed bad. Internally they stay separate, I just though to make the API easier to use. But I'm starting to think you may be right that it would be better to make `set_progress()` consistent with the internal split. Also, maybe `set_progress()` is a bad interface, and we should instead have `set_elapsed()` which takes a `Duration` and has enough precision with 128 bits.

> Everything should maybe wrap instead? The transition might be glitchy, but at least the animation won't get stuck in infinite mode.
> [...]
> I think this is acceptable given enough bits because no one is going to be repeating exactly some number of times greater than 1000.

Yes that was my reasoning too, I doubt any repeat count will go over, say 1 million to be conservative. Think of a UI animation repeating every 250ms on screen, forever while the game plays, over 1 year of gameplay (!) that's "only" 126 millions, so fits in 32 bits. Unlikely we ever hit the wrapping. Saturating or wrapping, probably doesn't matter.

> What do you think about this plan: use 16 bits of times_completed, keep f32 progress, make progress always use fract instead of clamping (aka be times_completed.progress), get rid of the times_completed API.

Sounds good, but keeping 32-bit `times_completed`, at least for now. With 16 bits only, my example above reaches capacity at 4.5 hours of game running, which is quite common to hit. If we really think we need to save 2 extra bytes we can review and nail down a precise wrap around behavior in a future change; in the meantime, I propose we just use 32 bits and push the problem away enough that we don't need to care.

> In terms of PRs, this one is getting too big for me so I'd like to get agreement on future plans, merge it, and then fix whatever the plan is + sequences + tracks in separate PRs which means things might be in a slightly broken state in the interim.

Sounds good.

> BTW, I think this solves https://github.com/djeedai/bevy_tweening/issues/8. The answer to what happens with PingPong asymmetric tracks is "compute the times_completed.progress for each track given that 1.0 is defined as the length of the longest track." This works fine with infinitely repeating tweens too since we only care about how long a single repetition is.

I don't think it fully does. What happens with the previous example of tweens A and B if tween B is 5 seconds long, and tween A is 1 second long and repeat 4 times (so total 4 seconds). What happens between t=4s and t=5s for tween A? And what happens at t=5s when reversing direction? I don't think there's a unique answer, but it's still unclear to me what's the least surprising behavior for users here.

djeedai: > > I'm not sure your model works because what if Tween A repeats infinitely?
>
> This is explicitly disallowed. I'm pretty sure I documented that somewhere. That's one limitation of Tracks.

That being said, we _could_ make that work by saying that the `Tracks` itself becomes of infinite duration/repeat, and that any child tweenable that doesn't repeat infinitely is some kind of "intro" that will only ever be played once at start. Not sure it's worth the hassle though.

SUPERCILEX: That's it! `set_elapsed` is the missing piece. It's objectively the better API because it's more fundamental — progress and times_completed can be derived from the combination of duration and elapsed.

This also leads me to believe number of completions for the repetition count is the wrong way to go. It should instead be a max duration. That's strictly more flexible since you can end an animation in a partially completed state. Again, it's also a more fundamental API because you can derive a max duration from a desired number of completions.

---

> I don't think it fully does. What happens with the previous example of tweens A and B if tween B is 5 seconds long, and tween A is 1 second long and repeat 4 times (so total 4 seconds). What happens between t=4s and t=5s for tween A? And what happens at t=5s when reversing direction? I don't think there's a unique answer, but it's still unclear to me what's the least surprising behavior for users here.

I agree that there isn't a unique answer, but I think if we get the fundamentals right, we'll be able to find solutions that maximize consistency. `set_elapsed` for example makes updating the progress of tracks trivially simple: call `set_elapsed` on each sub tween. Infinite tweens don't cause problems because they'll update the same as everyone else with the only difference being that they'll never finish.

On that note, I think we need to figure out infinite tweens once and for all. The core conflict lies in the fact that we want non-infinite tweens to run through their repetitions before we move on whereas infinite tweens doing the same makes no sense. I spent a few hours thinking about this and didn't really get anywhere: you end up with a bunch of contradictions or limitations if you do weird stuff with the infinities. Thus, the most consistent answer is likely the simplest: infinity means infinity. If you put something infinite in a sequence or track, then any non-zero number of repetitions is the same because you'll never actually repeat.

As for your example, I believe the most flexible answer is to say that tween A will stop and reversing directions will simply call the tween in the same order but with the direction reversed. If it turns out you wanted symmetry, you can create a sequence with Tween A and a delay of 1s afterwards. On the other hand, if we enforce symmetry directly in the tracks, then there's no way for people to get A and B starting at the same time but reversed which is why we have to let the user build symmetry for themselves if so desired. Perhaps if people want it, we could add a constructor for tracks that automatically makes every tracks symmetric by wrapping it in a sequence + delay, but the point is that the API is flexible enough for us to be able to do that in the future.

Ok, so here's the updated plan:
- Get rid of all notion of progress.
  - This (and times_completed) could be added back via an extension trait that only uses the public API of Tweenable if people want it.
- Get rid of the times_completed API.
- Get rid of the set_repeat_* methods because changing the repeat count of a live tween in a sequence for example wouldn't make any sense (our duration would be wrong).
- Add `RepeatCount::Until(Duration)`. Under the hood, `Finite(u32)` will just convert itself to `Until` once we know the tween duration.
- Unrelated but I just noticed this: get rid of `enabled` in *_completed_event methods since you basically always pass in true.
  - Add a `clear_completed_event` method to be consistent with the callback APIs.
- Add `set_elapsed` and `elapsed` methods.
- Make sequences and tracks simply wait for their tweens to complete before repeating (which means infinite tweens block their completion).
- Tracks always play their tweens in the same order when repeating (meaning creating symmetry is left to the user through a combination of sequence + delay).

djeedai: > This also leads me to believe number of completions for the repetition count is the wrong way to go. It should instead be a max duration. That's strictly more flexible since you can end an animation in a partially completed state. Again, it's also a more fundamental API because you can derive a max duration from a desired number of completions.

That looks like a very low level API not very practical for common use cases. Ending an animation in a partially completed state is probably not a very common use case, is it? Not objecting to having that low-level API, but probably need to retain the higher-level one too based on repeat count.

> I agree that there isn't a unique answer, but I think if we get the fundamentals right, we'll be able to find solutions that maximize consistency. set_elapsed for example makes updating the progress of tracks trivially simple: call set_elapsed on each sub tween. Infinite tweens don't cause problems because they'll update the same as everyone else with the only difference being that they'll never finish.

Very good point. That will cause issue to implement `Tweenable::duration()` though if we allow infinite tweenables inside a `Tracks`. Or maybe we need to return `Option<Duration>` there or an enum `Finite(Duration)` vs. `Infinite`?

> If you put something infinite in a sequence or track, then any non-zero number of repetitions is the same because you'll never actually repeat.

Yes, we can't repeat a sequence or tracks with an infinite child. We could maybe throw a warning in that case on creation, or return a `Result<>` somewhere to catch those misuses.

> As for your example, I believe the most flexible answer is to say that tween A will stop and reversing directions will simply call the tween in the same order but with the direction reversed. If it turns out you wanted symmetry, you can create a sequence with Tween A and a delay of 1s afterwards. On the other hand, if we enforce symmetry directly in the tracks, then there's no way for people to get A and B starting at the same time but reversed which is why we have to let the user build symmetry for themselves if so desired. Perhaps if people want it, we could add a constructor for tracks that automatically makes every tracks symmetric by wrapping it in a sequence + delay, but the point is that the API is flexible enough for us to be able to do that in the future.

All agreed.

> - Get rid of all notion of progress.
>   - This (and times_completed) could be added back via an extension trait that only uses the public API of Tweenable if people want it.
> - Get rid of the times_completed API.

Not sure I'd be as aggressive as to totally remove them, but yes at least implement them in terms of the more fundamental `set_elapsed()`, and avoid referring to those concepts internally in the implementation.

> - Get rid of the set_repeat_* methods because changing the repeat count of a live tween in a sequence for example wouldn't make any sense (our duration would be wrong).

I guess this was already not well defined, so removing probably makes sense. We can re-add later if we have a well-defined behavior for this.

> - Add RepeatCount::Until(Duration). Under the hood, Finite(u32) will just convert itself to Until once we know the tween duration.

Sounds good.

> - Unrelated but I just noticed this: get rid of enabled in *_completed_event methods since you basically always pass in true.
>   - Add a clear_completed_event method to be consistent with the callback APIs.

See other PR; this is just a different API proposal for the same thing. Not necessarily opposed to it, if there's a rationale to justify this one is better than the existing one. Not sure personally which one is better.

> - Make sequences and tracks simply wait for their tweens to complete before repeating (which means infinite tweens block their completion).

This is already the case for `Sequence`, but yes agreed for `Tracks`.

> - Tracks always play their tweens in the same order when repeating (meaning creating symmetry is left to the user through a combination of sequence + delay).

Yes.

SUPERCILEX: > need to retain the higher-level one too based on repeat count.

Agreed.

> Or maybe we need to return Option<Duration> there or an enum Finite(Duration) vs. Infinite?

I was thinking we could just return Duration::MAX, but I think I like the enum better.

> Yes, we can't repeat a sequence or tracks with an infinite child. We could maybe throw a warning in that case on creation, or return a Result<> somewhere to catch those misuses.

I'm not sure that's worth it because it does make sense to use infinite tweens. For example you could make a track for a tutorial that moves an object to where the player is supposed to focus and also wiggles the object. You could want the wiggle animation to go on forever while the movement animation only completes N times.

I guess we could crash if you use a repeat count greater than 1, but eh, it's not an invalid state.

> Not sure I'd be as aggressive as to totally remove them, but yes at least implement them in terms of the more fundamental set_elapsed(), and avoid referring to those concepts internally in the implementation.

In that case I'd like to add them back via an extension trait so as to not pollute the core API.

> I guess this was already not well defined, so removing probably makes sense. We can re-add later if we have a well-defined behavior for this.

👍

---

Woohoo, looks like we have a path forward! I'll try to get this PR fixed up today or sometime this week.

SUPERCILEX:
@djeedai Ok, I think this PR is ready to be merged! Stuff is still a bit half-baked, but I'll clean it up in the next PR.

src/tweenable.rs
```diff
@@ -488,22 +488,17 @@ impl<T> Tweenable<T> for Tween<T> {
...
-        if !self.is_looping() && self.clock.completed() {
+        if self.clock.state() == TweenState::Completed {
             return TweenState::Completed;
         }
 
         // Tick the animation clock
         let times_completed = self.clock.tick(delta);
-        self.times_completed += times_completed;
-        if times_completed & 1 != 0 && self.tweening_type == TweeningType::PingPong {
+        self.clock.record_completions(times_completed);
```
djeedai, line 497: It's a bit awkward that we do `clock.tick()`, get back a times completed, and re-inject that immediately into the clock. Shouldn't that recording be updated during `tick()` instead?

djeedai (changes requested):
A few minor things but should be good after that!

examples/sequence.rs
```diff
@@ -109,17 +109,28 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
-                end: pair[1] - center,
-            },
-        )
-        // Get an event after each segment
-        .with_completed_event(index as u64)
+        Tracks::new([
+            Tween::new(
+                EaseFunction::QuadraticInOut,
+                Duration::from_millis(250),
+                TransformRotationLens {
```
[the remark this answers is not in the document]
SUPERCILEX: TIL, done.

src/lib.rs
```diff
@@ -174,24 +165,47 @@ pub use tweenable::{
...
+mod plugin;
+mod tweenable;
+
+/// How many times to repeat a tween animation. See also: [`RepeatStrategy`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RepeatCount {
+    /// Run the animation N times.
+    Finite(u32),
+    /// Run the animation for some amount of time.
+    Until(Duration),
```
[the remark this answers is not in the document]
SUPERCILEX: Went with `For` because `During` felt a little weird.

src/tweenable.rs
SUPERCILEX: Agreed, but I'm just going to delete that code in the next PR so I'd rather not move it.

SUPERCILEX:
The CI error is fake: `note: D:\a\bevy_tweening\bevy_tweening\target\debug\examples\transform_translation.exe : fatal error LNK1180: insufficient disk space to complete link`

djeedai (approved):

SUPERCILEX:
@djeedai oh wait, are you going to ship these changes in the 0.8 compatible version? I'll have to speed up a bit then so we end up in a nice place. :)

djeedai:
> @djeedai oh wait, are you going to ship these changes in the 0.8 compatible version? I'll have to speed up a bit then so we end up in a nice place. :)

Yes I am, this is a very good feature I think, which if I remember has been asked on Discord. If we can get the progress update (#31) that would be nice too, but I'd like to get a new release supporting 0.8 as soon as possible, as several people asked for it and are waiting, and there's an upcoming Bevy jam too. Let me know if you think you can make it, or we can cut today-ish and we can make another release 0.5.1 later with some added features _etc._

SUPERCILEX:
Sounds good, I'll see what I'm able to get done. I just want to make sure we aren't rushing.

Issue #16, opened 2022-04-24
Make `Sequence` loopable


djeedai:
This is a sub-issue of #8. Opening this separately because for `Sequence` the problem is I believe well-defined, and work can move forward. This is not the case for `Tracks` though, so splitting the issue in two for clarity.

Allow `Sequence` to be loopable both with `TweeningType::Loop` (play all child tweenables once, then restart at first one) and with `TweeningType::PingPong` (play all child tweenables once, then restart from last back to first, then restart from first to last, _etc._).

PR #21, opened 2022-05-08
Fix questionable use of floats and re-work all ticking logic 

SUPERCILEX:
What's broken?
- Float usage inconsistently used f64 or f32. I changed everything to be f32 since precision isn't really the priority in games.
- Float rounding issues were present in several places. For example, the AnimClock never reset the elapsed time leading to a continual loss of precision as the animation progressed.
- Sequences did not handle deltas that crossed animation boundaries. This meant that if a one-hour delta was received, only one animation would complete instead of all of them.
- I tried to simplify all the other logic in the process while improving performance.

### DIFF
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -1,5 +1,7 @@
-use bevy::prelude::*;
+use std::cmp::min;
 use std::time::Duration;
+
+use bevy::prelude::*;
 
 use crate::{EaseMethod, Lens, TweeningDirection, TweeningType};
 
@@ -45,7 +47,7 @@ pub struct TweenCompleted {
 #[derive(Debug, Default, Clone, Copy)]
 struct AnimClock {
     elapsed: Duration,
-    total: Duration,
+    duration: Duration,
     is_looping: bool,
 }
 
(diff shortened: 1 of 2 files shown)


src/tweenable.rs
```diff
@@ -45,60 +47,52 @@ pub struct TweenCompleted {
...
+            1
+        }
     }
 
-    fn set_progress(&mut self, progress: f32) -> u32 {
-        let progress = progress.max(0.);
-        let times_completed = progress as u32;
+    fn set_progress(&mut self, progress: f32) {
         let progress = if self.is_looping {
             progress.fract()
```
djeedai, line 82: Because of the deleted `.max(0.)`, this may set `progress` to a negative value since `(-3.6_f32).fract() == -0.6_f32`, which will make `Duration::mul_f32()` panic.

In general the interface for `set_progress()` shouldn't allow negative values, that makes little sense unless the animation is infinitely looping. So it's better to just avoid it completely for the sake of simplicity and clarity of the API.

src/tweenable.rs
```diff
@@ -561,72 +551,122 @@ impl<T> Tweenable<T> for Sequence<T> {
...
+        if progress < 1e-5 {
+            self.rewind();
+            return;
+        } else if progress > 1. - 1e-5 {
+            self.elapsed = self.duration;
+            self.index = self.tweens.len();
+            return;
+        }
+
+        self.elapsed = self.duration.mul_f32(progress.clamp(0., 1.));
```
djeedai, line 564: The interface for `set_progress()` should accept values > `1.`, which allows jumping to iteration N of a repeating animation. Looping sequences are not yet implemented but there's work on-going for them (#16).

src/tweenable.rs
```diff
@@ -561,72 +551,122 @@ impl<T> Tweenable<T> for Sequence<T> {
...
+            self.index += 1;
+            if self.index == len {
+                // We've reached the end so we don't care about the remaining delta.
+                break;
+            }
+
+            let tween_duration = tween.duration();
+
+            let full_completions =
+                (tween.times_completed() - prev_completions - 1) * tween_duration;
```
djeedai, line 642: Note that this seems to attempt to handle a case that is currently not supported, namely having looping tweenables as part of a `Sequence`. Currently `full_completions` will always be zero. Also, if a tweenable is looping, it currently never return `TweenState::Completed` since looping tweenables are of infinite duration by definition.

djeedai (changes requested):
Thanks for the contribution! There's a bunch of good things in there, but also unfortunately a few other ones that are either arguable or conflict with on-going works toward making `Sequence` (and possibly `Tracks`) loopable, and therefore would be reverted as soon as that work lands. I'm also not very clear on where are the bugs being fixed.

I'd suggest splitting the changes into smaller chunks, separating the targeted bug fix(es), the optimization(s), and the clean-ups around `f32` / `f64` mix. We can chat on Discord if you want.

src/tweenable.rs
```diff
@@ -45,60 +47,52 @@ pub struct TweenCompleted {
...
-        let progress = new_elapsed.as_secs_f64() / self.total.as_secs_f64();
-        let times_completed = progress as u32;
-        let progress = if self.is_looping {
-            progress.fract()
+            self.elapsed = Duration::from_secs_f32(elapsed % duration);
+            ((elapsed / duration) + 1e-5).floor() as u32
         } else {
-            progress.min(1.)
-        };
-        self.elapsed = self.total.mul_f64(progress);
```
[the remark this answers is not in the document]
SUPERCILEX: Well that's embarrassing, I totally missed that. :sweat: But! Using your `as_nanos` suggestion I managed to almost double performance while slightly improving precision loss: https://github.com/djeedai/bevy_tweening/pull/24

PR #22, opened 2022-05-08
Add speed modifier

SUPERCILEX:
It needs to be built-in because otherwise you have to completely recreate the animation to change the speed.

I didn't add a `get_speed` because that's a meaningless question for tweenables like `Sequence`. I also updated some of the examples to let people vary the speed... I can do all of them if desired.

### DIFF
--- a/examples/transform_rotation.rs
+++ b/examples/transform_rotation.rs
@@ -1,7 +1,9 @@
 use bevy::prelude::*;
+use bevy_egui::{egui, EguiContext, EguiPlugin};
+
 use bevy_tweening::{lens::*, *};
 
-fn main() -> Result<(), Box<dyn std::error::Error>> {
+fn main() {
     App::default()
         .insert_resource(WindowDescriptor {
             title: "TransformRotationLens".to_string(),
@@ -12,10 +14,27 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         })
         .add_plugins(DefaultPlugins)
         .add_plugin(TweeningPlugin)
+        .add_plugin(EguiPlugin)
         .add_startup_system(setup)
+        .init_resource::<Options>()
+        .add_system(options_panel)
+        .add_system(update_animation_speed)
         .run();
+}
 
-    Ok(())
+#[derive(Copy, Clone, PartialEq)]
+struct Options {
+    open: bool,
+    speed: f32,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            open: true,
+            speed: 1.,
+        }
+    }
 }
 
 fn setup(mut commands: Commands) {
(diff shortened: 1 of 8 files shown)


examples/transform_rotation.rs
```diff
@@ -96,3 +115,33 @@ fn setup(mut commands: Commands) {
         }
     }
 }
+
+fn options_panel(mut egui_context: ResMut<EguiContext>, mut options: ResMut<Options>) {
```
djeedai, line 119: Can't this be done with `bevy-inspector-egui` instead of adding an extra dependency?

src/tweenable.rs
```diff
@@ -119,6 +121,13 @@ pub trait Tweenable<T>: Send + Sync {
     /// reach back the same state in this case is the double of the returned value.
     fn duration(&self) -> Duration;
 
+    /// Set the animation speed. Defaults to 1.
+    ///
+    /// Speeds greater than 1 slow down time. That is, a speed of 10 means the animation will
+    /// take 10 times longer to complete whereas a speed of 0.5 means the animation will complete
+    /// twice as fast.
```
djeedai, line 128: This is really counter-intuitive. Higher speeds should play the animation faster, not slower.

djeedai:
Although adding variable playback speed is a good feature to add, I'm skeptical about the approach of integrating that into each tweenable. I'd rather this concept stay at the level of the `Animator`, which can simply pass a modified time to `Tweenable::tick()` based on the game time and animator's own speed multiplier. This keeps the change local to the `Animator`, and make it work for all kind of tweenables present and future. This also saves having to store an extra field in each child `Tween`.

SUPERCILEX:
Sounds good, done.

examples/transform_rotation.rs
SUPERCILEX: Didn't know you could do that, pretty cool!

src/lib.rs
```diff
@@ -513,6 +561,21 @@ impl<T: Asset> AssetAnimator<T> {
         }
     }
 
+    /// Ticks the tween, if present. See [`Tweenable::tick`] for details.
+    pub fn tick(
+        &mut self,
+        delta: Duration,
+        target: &mut T,
+        entity: Entity,
+        event_writer: &mut EventWriter<TweenCompleted>,
```
djeedai, line 570: This forces a change on the `EventWriter` (when the caller derefs the `ResMut`) even when there's no animation. This should probably take a `&mut ResMut` instead. Same for `Animator<T>::tick()`.

djeedai (changes requested):
Almost all good. Can you please fix the small `RefMut` comment, and update the `CHANGELOG.md`? Thanks!

src/lib.rs
SUPERCILEX: Hmmm, I don't understand. Where are ResMuts involved? This is an event writer that's being passed in by reference which means it shouldn't be getting derefed.

djeedai: Hum sorry my bad I misread indeed, I thought about the general issue and didn't check that `EventWriter` is passed by value in the system. Nevermind, please ignore that.

djeedai (approved):

src/lib.rs
SUPERCILEX: No worries lol :)

PR #24, opened 2022-05-16
Fix loss of precision in AnimClock and improve performance

SUPERCILEX:
Old:

1s 320ms
Left:  140292640
Right: 140292630

New:

759ms
Success

### DIFF
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -45,7 +45,7 @@ pub struct TweenCompleted {
 #[derive(Debug, Default, Clone, Copy)]
 struct AnimClock {
     elapsed: Duration,
-    total: Duration,
+    duration: Duration,
     is_looping: bool,
 }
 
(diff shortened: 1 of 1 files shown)


SUPERCILEX:
Oh, for whatever reason now I'm seeing ~1s time for the old version. The new one is still faster, but not by as much I guess.

djeedai:
If you want stable results you need to `bench` it. This test is not warming the codepath before measuring, so will have variations per run. An ideal perf test runs the code first a few times so that all caches are hot and the CPU frequency scaling if any is already in power/perf mode (and ideally disables low-power and frequency scaling, but that's not always possible). That's probably good enough though given the large difference before/after.

SUPERCILEX:
Yeah, I'm being very lazy. I mainly just wanted to check that using u128s wasn't slower than the old stuff.

djeedai:
I'm going to merge as is for now, but I think the removal of `times_completed` will have to be reverted for #16 to work. Anyway, we'll see.

djeedai (approved):

SUPERCILEX:
Sounds good! I don't have time this week, but I'll start working on your ideas for https://github.com/djeedai/bevy_tweening/pull/19 (unless you get to it first :smile:) which will let me see what's still applicable in https://github.com/djeedai/bevy_tweening/pull/21.

PR #25, opened 2022-05-21
Add TransformLens and IsometricTransformLens types

istanbul-not-constantinople:
#### This allows easier tweening of different Transform fields together:
```rust
Tween::new(
    EaseFunction::QuadraticInOut,
    TweeningType::Once,
    Duration::from_secs(1)
    IsometricTransformLens {
        start_position: Vec3::ZERO,
        end_position: Vec3::new(0., 5., 0.),
        start_rotation: Quat::IDENTITY,
        end_rotation: Quat::from_axis_angle(Vec3::new(0., 1., 0.), 3.),
    }
).then(Tween::new(
    EaseFunction::QuadraticInOut,
    TweeningType::PingPong,
    Duration::from_secs_f32(0.5),
    TransformLens {
        start: Transform::identity(),
        end: Transform {
            translation: Vec3::new(5., 0., 0.),
            rotation: Quat::from_axis_angle(Vec3::new(1., 0., 0.), 3.),
            scale: Vec3::splat(2.),
        }
    }
));
```

### DIFF
--- a/src/lens.rs
+++ b/src/lens.rs
@@ -168,7 +168,7 @@ impl Lens<Transform> for TransformLens {
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub struct IsometricTransformLens {
     /// Start value of the postition.
-    pub start_postition: Vec3,
+    pub start_position: Vec3,
     /// End value of the position.
     pub end_position: Vec3,
     /// Start value of the rotation.
--- a/src/lens.rs
+++ b/src/lens.rs
@@ -98,6 +98,92 @@ impl Lens<Text> for TextColorLens {
     }
 }
 
+/// A lens to manipulate a [`Transform`] component.
+/// 
+/// Each of the fields have their own lenses which should be used independently
+/// where possible:
+/// - [`TransformPositionLens`]
+/// - [`TransformScaleLens`]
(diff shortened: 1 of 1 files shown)


djeedai (commented):
Hi- Thanks for opening a PR, contributions are very much welcome. In this instance though, I think this goes against the design of the library. One of the central ideas of the `Lens` approach is that it's very easily extensible from user code without having to modify the library and maintain many lenses internally for all the various use cases. I'm skeptical this extra lens provides a large enough added value to justify having to maintain it internally, and I don't see any drawback in leaving it to the user to write it. If anything, the internal lenses are more like examples of how to use lenses; it's very much expected users will write their own lenses to tailor them to their use case, like you did here. Why did you want to integrate that lens?

istanbul-not-constantinople:
In the project in working on, all my tweens are isometric transforms which I'm using just to move and rotate different entities to dynamically assigned targets, so that was the use-case I had in mind.

> If anything, the internal lenses are more like examples of how to use lenses

Yeah. I can see that now and I agree with you there. The implementation is less than 50 lines together so I think it's small enough to incorporate into downstream projects without much hassle. Closing for now; if a better reason ever arises (e.g. Bevy implements their own `Isometry` struct) this could be re-opened.

PR #26, opened 2022-05-29
Slightly modernize codebase

SUPERCILEX:
Apply a bunch of clippy suggestions.

### DIFF
--- a/examples/colormaterial_color.rs
+++ b/examples/colormaterial_color.rs
@@ -5,21 +5,18 @@ use bevy::{
 use bevy_tweening::{lens::*, *};
 use std::time::Duration;
 
-fn main() -> Result<(), Box<dyn std::error::Error>> {
+fn main() {
     App::default()
         .insert_resource(WindowDescriptor {
             title: "ColorMaterialColorLens".to_string(),
             width: 1200.,
             height: 600.,
-            present_mode: bevy::window::PresentMode::Fifo, // vsync
-            ..Default::default()
+            ..default()
         })
         .add_plugins(DefaultPlugins)
         .add_plugin(TweeningPlugin)
         .add_startup_system(setup)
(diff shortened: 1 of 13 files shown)


examples/colormaterial_color.rs
```diff
@@ -5,21 +5,18 @@ use bevy::{
...
-fn main() -> Result<(), Box<dyn std::error::Error>> {
+fn main() {
     App::default()
         .insert_resource(WindowDescriptor {
             title: "ColorMaterialColorLens".to_string(),
             width: 1200.,
             height: 600.,
-            present_mode: bevy::window::PresentMode::Fifo, // vsync
-            ..Default::default()
+            ..default()
```
djeedai, line 14: Why remove vsync? It is indeed the default, but it's not very clear (I had to dig through the Bevy sources) and is not documented as such.

SUPERCILEX: I thought it was repetitive since it's the default, but I don't care either way. Added back.

src/lib.rs
```diff
@@ -223,10 +223,11 @@ pub enum EaseMethod {
 }
 
 impl EaseMethod {
+    #[must_use]
     fn sample(self, x: f32) -> f32 {
         match self {
             EaseMethod::EaseFunction(function) => x.calc(function),
-            EaseMethod::Linear => x,
+            Self::Linear => x,
```
djeedai, line 230: This is a bit awkward, why change only that one and not the ones above and below? 😅

djeedai (approved):

src/lib.rs
SUPERCILEX: Good question, I guess that's a clippy bug. Fixed.

djeedai (approved):

PR #30, opened 2022-07-09
Make completed event API consistent

SUPERCILEX:


djeedai:
Consistent with what? This just changes the API for a different one. What's the rationale?

SUPERCILEX:
Consistent with the completed callback API which is better because it makes invalid states unrepresentable. The current API lets you say `false, something` where the something is garbage.

djeedai:
Good point. All for a more robust API.

djeedai (approved):

PR #34, opened 2022-08-01
Update to bevy 0.8

outergod:
bevy 0.8 has just been released and I would love to use bevy_tweening with it. I've successfully been using these changes with a local copy. `bevy-inspector-egui` does not have a compatible release yet, unfortunately.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,10 @@ bevy_ui = ["bevy/bevy_ui", "bevy/bevy_text", "bevy/bevy_render"]
 
 [dependencies]
 interpolation = "0.2"
-bevy = { version = "0.7", default-features = false }
+bevy = { version = "0.8", default-features = false, features = ["bevy_asset"] }
 
-[dev-dependencies]
-bevy-inspector-egui = "0.10"
+# [dev-dependencies]
+# bevy-inspector-egui = "0.11"
 
 [[example]]
 name = "menu"
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -18,7 +18,7 @@ criterion = { version = "0.3", features = ["html_reports"] }
 bevy_tweening = { path = "../" }
 
 [dependencies.bevy]
-version = "0.6"
+version = "0.8"
 default-features = false
 features = [ "render" ]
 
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -21,10 +21,10 @@ bevy_ui = ["bevy/bevy_ui", "bevy/bevy_text", "bevy/bevy_render"]
 
 [dependencies]
 interpolation = "0.2"
-bevy = { version = "0.7", default-features = false }
+bevy = { version = "0.8", default-features = false, features = ["bevy_asset"] }
```
SUPERCILEX, line 24: `bevy_asset` should have its own separate feature so users aren't forced to enable it. Probs not worth blocking the upgrade for this so created

djeedai:
I sync'ed with the `bevy-inspector-egui` maintainer over the week-end and I'm waiting for them to upgrade. I was told this should be soon. I have a local upgrade of `bevy_tweening` ready after that, and we'll snap a `0.5.0` targeting Bevy 0.8 and including with all the cool stuffs @SUPERCILEX added since `0.4.0`. I want to avoid having to disable `bevy-inspector-egui` then re-enable it later and have to make 2 releases.

djeedai:
Blocked on jakobhellermann/bevy-inspector-egui#75

outergod:
https://github.com/jakobhellermann/bevy-inspector-egui/issues/75 is resolved and closed now.

djeedai:
> [jakobhellermann/bevy-inspector-egui#75](https://github.com/jakobhellermann/bevy-inspector-egui/issues/75) is resolved and closed now.

Yes I saw that, thanks @outergod! I'm probably going to push my own change for 0.8 support instead of this PR, because I'm adding the `bevy_asset` feature on the fly, and several other small changes needed in the README etc. to keep things consistent.

SUPERCILEX:
@djeedai I've been thinking about this and I believe the right move is to branch off of the 0.4 tag and *only* upgrade Bevy. That branch can be kept around or deleted so only the 0.5 tag stays. Anyway, the two key benefits of releasing 0.5 that only bumps bevy support are:
- Users don't have to try and fix compilation issues across both bevy and tweening.
- We get more time to release a fully polished upgrade.

If you're fine waiting a week or two to release, I think we can get everything implemented, but it's not going to be done by tomorrow or Friday.

djeedai:
I've upgraded myself and released 0.5.0 because this was a bit complicated with some revert _etc._ in the middle. Thanks @outergod for the contribution though!

PR #37, opened 2022-08-03
Improve testing framework

SUPERCILEX:
The current tests are quite painful to work with since it's very difficult to tell what the expected values should actually be. This PR flips the testing around by passing in a sequence of expected values that the test framework will validate.

Note: depends on https://github.com/djeedai/bevy_tweening/pull/19.

### DIFF
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -45,7 +45,7 @@ pub struct TweenCompleted {
 #[derive(Debug, Default, Clone, Copy)]
 struct AnimClock {
     elapsed: Duration,
-    total: Duration,
+    duration: Duration,
     is_looping: bool,
 }
 
@@ -53,52 +53,44 @@ impl AnimClock {
     fn new(duration: Duration, is_looping: bool) -> Self {
         AnimClock {
             elapsed: Duration::ZERO,
-            total: duration,
+            duration,
             is_looping,
         }
     }
 
-    #[allow(dead_code)] // TEMP
-    fn elapsed(&self) -> Duration {
-        self.elapsed
-    }
-
-    fn total(&self) -> Duration {
-        self.total
-    }
-
     fn tick(&mut self, duration: Duration) -> u32 {
-        let new_elapsed = self.elapsed.saturating_add(duration);
-        let progress = new_elapsed.as_secs_f64() / self.total.as_secs_f64();
-        let times_completed = progress as u32;
-        let progress = if self.is_looping {
-            progress.fract()
+        self.elapsed = self.elapsed.saturating_add(duration);
+
+        if self.elapsed < self.duration {
+            0
+        } else if self.is_looping {
+            let elapsed = self.elapsed.as_nanos();
+            let duration = self.duration.as_nanos();
+
+            self.elapsed = Duration::from_nanos((elapsed % duration) as u64);
+            (elapsed / duration) as u32
         } else {
-            progress.min(1.)
-        };
-        self.elapsed = self.total.mul_f64(progress);
-        times_completed
+            self.elapsed = self.duration;
+            1
+        }
     }
 
-    fn set_progress(&mut self, progress: f32) -> u32 {
-        let progress = progress.max(0.);
-        let times_completed = progress as u32;
+    fn set_progress(&mut self, progress: f32) {
         let progress = if self.is_looping {
-            progress.fract()
+            progress.max(0.).fract()
         } else {
-            progress.min(1.)
+            progress.clamp(0., 1.)
         };
-        self.elapsed = self.total.mul_f32(progress);
-        times_completed
+
+        self.elapsed = self.duration.mul_f32(progress);
     }
 
     fn progress(&self) -> f32 {
-        //self.elapsed.div_duration_f32(self.total) // TODO: unstable
-        (self.elapsed.as_secs_f64() / self.total.as_secs_f64()) as f32
+        self.elapsed.as_secs_f32() / self.duration.as_secs_f32()
     }
 
     fn completed(&self) -> bool {
-        self.elapsed >= self.total
+        self.elapsed >= self.duration
     }
 
     fn reset(&mut self) {
@@ -408,7 +400,7 @@ impl<T> Tween<T> {
 
 impl<T> Tweenable<T> for Tween<T> {
     fn duration(&self) -> Duration {
-        self.clock.total()
+        self.clock.duration
     }
 
     fn is_looping(&self) -> bool {
@@ -806,6 +798,36 @@ mod tests {
     struct CallbackMonitor {
         invoke_count: u64,
         last_reported_count: u32,
+    }
+
+    #[test]
+    fn anim_clock_precision() {
+        let duration = Duration::from_millis(1);
+        let mut clock = AnimClock::new(duration, true);
+
+        let test_ticks = [
+            Duration::from_micros(123),
+            Duration::from_millis(1),
+            Duration::from_secs_f32(1. / 24.),
+            Duration::from_secs_f32(1. / 30.),
+            Duration::from_secs_f32(1. / 60.),
+            Duration::from_secs_f32(1. / 120.),
+            Duration::from_secs_f32(1. / 144.),
+            Duration::from_secs_f32(1. / 240.),
+        ];
+
+        let mut times_completed = 0;
+        let mut total_duration = Duration::ZERO;
+        for i in 0..10_000_000 {
+            let tick = test_ticks[i % test_ticks.len()];
+            times_completed += clock.tick(tick);
+            total_duration += tick;
+        }
+
+        assert_eq!(
+            (total_duration.as_secs_f64() / duration.as_secs_f64()) as u32,
+            times_completed
+        );
     }
 
     /// Test ticking of a single tween in isolation.
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -1,7 +1,7 @@
 use bevy::prelude::*;
 use std::time::Duration;
 
-use crate::{EaseMethod, Lens, TweeningDirection, TweeningType};
+use crate::{EaseMethod, Lens, RepeatCount, RepeatStrategy, TweeningDirection};
 
 /// Playback state of a [`Tweenable`].
 ///
@@ -42,7 +42,7 @@ pub struct TweenCompleted {
     pub user_data: u64,
 }
 
-#[derive(Debug, Default, Clone, Copy)]
+#[derive(Debug)]
 struct AnimClock {
     elapsed: Duration,
     duration: Duration,
@@ -221,7 +221,8 @@ pub struct Tween<T> {
     ease_function: EaseMethod,
     clock: AnimClock,
     times_completed: u32,
-    tweening_type: TweeningType,
+    count: RepeatCount,
+    strategy: RepeatStrategy,
     direction: TweeningDirection,
     lens: Box<dyn Lens<T> + Send + Sync + 'static>,
     on_completed: Option<Box<CompletedCallback<T>>>,
@@ -238,7 +239,6 @@ impl<T: 'static> Tween<T> {
     /// # use std::time::Duration;
     /// let tween1 = Tween::new(
     ///     EaseFunction::QuadraticInOut,
-    ///     TweeningType::Once,
     ///     Duration::from_secs_f32(1.0),
     ///     TransformPositionLens {
     ///         start: Vec3::ZERO,
@@ -247,7 +247,6 @@ impl<T: 'static> Tween<T> {
     /// );
     /// let tween2 = Tween::new(
     ///     EaseFunction::QuadraticInOut,
-    ///     TweeningType::Once,
     ///     Duration::from_secs_f32(1.0),
     ///     TransformRotationLens {
     ///         start: Quat::IDENTITY,
@@ -271,7 +270,6 @@ impl<T> Tween<T> {
     /// # use std::time::Duration;
     /// let tween = Tween::new(
     ///     EaseFunction::QuadraticInOut,
-    ///     TweeningType::Once,
     ///     Duration::from_secs_f32(1.0),
     ///     TransformPositionLens {
     ///         start: Vec3::ZERO,
@@ -279,20 +277,16 @@ impl<T> Tween<T> {
     ///     },
     /// );
     /// ```
-    pub fn new<L>(
-        ease_function: impl Into<EaseMethod>,
-        tweening_type: TweeningType,
-        duration: Duration,
-        lens: L,
-    ) -> Self
+    pub fn new<L>(ease_function: impl Into<EaseMethod>, duration: Duration, lens: L) -> Self
     where
         L: Lens<T> + Send + Sync + 'static,
     {
         Tween {
             ease_function: ease_function.into(),
-            clock: AnimClock::new(duration, tweening_type != TweeningType::Once),
+            clock: AnimClock::new(duration, RepeatCount::default() != RepeatCount::Finite(1)),
             times_completed: 0,
-            tweening_type,
+            count: default(),
+            strategy: default(),
             direction: TweeningDirection::Forward,
             lens: Box::new(lens),
             on_completed: None,
@@ -313,7 +307,6 @@ impl<T> Tween<T> {
     /// let tween = Tween::new(
     ///     // [...]
     /// #    EaseFunction::QuadraticInOut,
-    /// #    TweeningType::Once,
     /// #    Duration::from_secs_f32(1.0),
     /// #    TransformPositionLens {
     /// #        start: Vec3::ZERO,
@@ -356,6 +349,29 @@ impl<T> Tween<T> {
     /// See [`Tween::set_direction()`].
     pub fn with_direction(mut self, direction: TweeningDirection) -> Self {
         self.direction = direction;
+        self
+    }
+
+    /// TODO
+    pub fn set_repeat_count(&mut self, count: RepeatCount) {
+        self.count = count;
+        self.clock.is_looping = count != RepeatCount::Finite(1);
+    }
+
+    /// TODO
+    pub fn with_repeat_count(mut self, count: RepeatCount) -> Self {
+        self.set_repeat_count(count);
+        self
+    }
+
+    /// TODO
+    pub fn set_repeat_strategy(&mut self, strategy: RepeatStrategy) {
+        self.strategy = strategy;
+    }
+
+    /// TODO
+    pub fn with_repeat_strategy(mut self, strategy: RepeatStrategy) -> Self {
+        self.strategy = strategy;
         self
     }
 
@@ -404,7 +420,11 @@ impl<T> Tweenable<T> for Tween<T> {
     }
 
     fn is_looping(&self) -> bool {
-        self.tweening_type != TweeningType::Once
+        match self.count {
+            RepeatCount::Finite(times) if times == 1 => false,
+            RepeatCount::Finite(times) => self.times_completed < times,
+            RepeatCount::Infinite => true,
+        }
     }
 
     fn set_progress(&mut self, progress: f32) {
@@ -429,14 +449,9 @@ impl<T> Tweenable<T> for Tween<T> {
         // Tick the animation clock
         let times_completed = self.clock.tick(delta);
         self.times_completed += times_completed;
-        if times_completed & 1 != 0 && self.tweening_type == TweeningType::PingPong {
+        if self.strategy == RepeatStrategy::Bounce && times_completed & 1 != 0 {
             self.direction = !self.direction;
         }
-        let state = if self.is_looping() || times_completed == 0 {
-            TweenState::Active
-        } else {
-            TweenState::Completed
-        };
         let progress = self.clock.progress();
 
         // Apply the lens, even if the animation finished, to ensure the state is consistent
@@ -460,7 +475,11 @@ impl<T> Tweenable<T> for Tween<T> {
             }
         }
 
-        state
+        if self.is_looping() || self.times_completed == 0 {
+            TweenState::Active
+        } else {
+            TweenState::Completed
+        }
     }
 
     fn times_completed(&self) -> u32 {
@@ -834,27 +853,29 @@ mod tests {
     #[test]
     fn tween_tick() {
         for tweening_direction in &[TweeningDirection::Forward, TweeningDirection::Backward] {
-            for tweening_type in &[
-                TweeningType::Once,
-                TweeningType::Loop,
-                TweeningType::PingPong,
+            for (count, strategy) in &[
+                (RepeatCount::Finite(1), RepeatStrategy::default()),
+                (RepeatCount::Infinite, RepeatStrategy::Teleport),
+                (RepeatCount::Finite(2), RepeatStrategy::Teleport),
+                (RepeatCount::Infinite, RepeatStrategy::Bounce),
+                (RepeatCount::Finite(2), RepeatStrategy::Bounce),
             ] {
                 println!(
-                    "TweeningType: type={:?} dir={:?}",
-                    tweening_type, tweening_direction
+                    "TweeningType: count={count:?} strategy={strategy:?} dir={tweening_direction:?}",
                 );
 
                 // Create a linear tween over 1 second
                 let mut tween = Tween::new(
                     EaseMethod::Linear,
-                    *tweening_type,
                     Duration::from_secs_f32(1.0),
                     TransformPositionLens {
                         start: Vec3::ZERO,
                         end: Vec3::ONE,
                     },
                 )
-                .with_direction(*tweening_direction);
+                .with_direction(*tweening_direction)
+                .with_repeat_count(*count)
+                .with_repeat_strategy(*strategy);
                 assert_eq!(tween.direction(), *tweening_direction);
                 assert!(tween.on_completed.is_none());
                 assert!(tween.event_data.is_none());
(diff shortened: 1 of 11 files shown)


SUPERCILEX:
@djeedai This PR is ready and is already showing its value by making it super easy to see that a bunch of stuff is broken (the transforms, completion counts, progress). I didn't bother converting the sequence/track/delay tests because I'll do those in a pre-PR before rewriting the implementations.

src/tweenable.rs
```diff
@@ -862,6 +869,20 @@ mod tests {
         (a - b).abs() < tol
     }
 
+    macro_rules! create_event_reader_writer {
```
djeedai, line 872: Why a macro? Can we use a regular function here?

src/tweenable.rs
```diff
@@ -894,264 +915,371 @@ mod tests {
...
+        tween.rewind();
+        assert!(abs_diff_eq(tween.progress(), 0., 1e-5));
+        assert_eq!(
+            tween.tick(
+                Duration::ZERO,
+                &mut transform,
+                Entity::from_raw(0),
+                &mut event_writer.get_mut(&mut world),
+            ),
+            TweenState::Active
```
djeedai, line 956: Is that still true if the tween previously completed? Here after the first `tick()` it's still active no? So there's little chance that will change with a `rewind()`. The test is still good, but maybe we'd need another one to make sure `rewind()` will un-complete a tween?

src/tweenable.rs
```diff
@@ -894,264 +915,371 @@ mod tests {
...
+        assert!(transform.translation.abs_diff_eq(Vec3::ZERO, 1e-5));
+        assert_eq!(tween.times_completed(), 0);
+    }
+
+    fn validate_tween(
+        mut tween: Tween<Transform>,
+        expected_values: impl Iterator<
+            Item = (
+                Duration,
+                (f32, u32, u32, TweeningDirection, TweenState, Transform),
```
djeedai, line 967: Can we use a named type? This list is getting out of hands, makes it hard to remember what is what.

src/tweenable.rs
```diff
@@ -894,264 +915,371 @@ mod tests {
...
+            ),
+        >,
+    ) {
+        dbg!(&tween.clock);
```
djeedai, line 971: Delete?

src/tweenable.rs
```diff
@@ -894,264 +915,371 @@ mod tests {
...
+        ) in expected_values
+        {
+            // Tick the tween
+            let actual_state = tween.tick(
+                delta,
+                &mut transform,
+                dummy_entity,
+                &mut event_writer.get_mut(&mut world),
+            );
+            dbg!(&tween.clock);
```
djeedai, line 1004: Delete?

src/tweenable.rs
```diff
@@ -74,10 +74,14 @@ pub enum TweenState {
 /// complete cycle start -> end -> start counts as 2 iterations and raises 2
 /// events (one when reaching the end, one when reaching back the start).
 ///
+/// If the tick delta is high enough that multiple completions have occurred in
```
djeedai, line 77: Maybe we should add a `completion_count` parameter to the even then, to report how many times a loop was completed during a single `tick()`?

src/tweenable.rs
```diff
@@ -894,264 +915,371 @@ mod tests {
...
+            [
+                TweeningDirection::Forward,
+                TweeningDirection::Backward,
+                TweeningDirection::Backward,
+                TweeningDirection::Forward,
+            ],
+            repeat(TweenState::Active),
+            [
+                Transform::from_translation(Vec3::splat(0.6)),
+                Transform::from_translation(Vec3::splat(-0.2)),
```
djeedai, line 1234: How is it ever possible that a tween animating a transform between x=0 and x=1 produces a negative x value?!

djeedai (changes requested):

src/tweenable.rs
SUPERCILEX: Agreed, I'll look into this later today. I think it should be possible to declare a type with a bunch of iterators that validate_tween smooshes together into a single iterator.

src/tweenable.rs
SUPERCILEX: Added a test for it.

src/tweenable.rs
SUPERCILEX: I think that's a reasonable improvement, but I'd personally rather not add anything until someone comes along with an actual use case. Otherwise it's just cruft that needs to be maintained without any clear users.

src/tweenable.rs
SUPERCILEX: Because `main` is broken. :) Progress currently increases forever so when we pong you get `1. - 1.2`.

I'm actually super happy that you spotted this because it was broken in #19 but the tests were too complicated for us to even notice.

djeedai: I'm now in favor of progress being just this thing always in `[0-1]` that only tells you how far along you are on a single loop. Anything else, especially since it's (or will be) a second-class concept, is too complicated I think and not valuable anymore.

src/tweenable.rs
SUPERCILEX: Done! The generics are pretty ugly, but at least the tests are clear.

src/tweenable.rs
SUPERCILEX: I'm personally still in favor of not including a progress API at all since it's not that hard to implement yourself with whatever semantics you like. That said, this won't be used internally so it doesn't matter too much and I'm fine implementing it either way.

SUPERCILEX: Though that's discussionfor #38

SUPERCILEX:
@djeedai This one should be ready for another round of review. It'd also be nice to get a decision on #44 since #38 will need to update Animator APIs otherwise.

src/tweenable.rs
```diff
@@ -862,6 +869,22 @@ mod tests {
         (a - b).abs() < tol
     }
 
+    fn create_event_reader_writer<'w, 's>() -> (
+        World,
+        SystemState<EventReader<'w, 's, TweenCompleted>>,
+        SystemState<EventWriter<'w, 's, TweenCompleted>>,
+    ) {
+        let mut world = World::new();
+        world.insert_resource(Events::<TweenCompleted>::default());
```
djeedai, line 878: nit:
```suggestion
        world.init_resource::<Events::<TweenCompleted>>();
```

src/tweenable.rs
```diff
@@ -894,264 +917,399 @@ mod tests {
...
-                assert_eq!(actual_state, TweenState::Active);
-                let expected_translation = if tweening_direction.is_backward() {
-                    Vec3::ONE
-                } else {
-                    Vec3::ZERO
-                };
-                assert!(transform
+        tween.set_progress(starting_progress);
+        tween.tick(
+            Duration::ZERO,
```
djeedai, line 941: nit: Add comment that this is to "apply" the `set_progress()` change?

src/tweenable.rs
```diff
@@ -894,264 +917,399 @@ mod tests {
...
+            &mut transform,
+            Entity::from_raw(0),
+            &mut event_writer.get_mut(&mut world),
+        );
+        // TODO(https://github.com/djeedai/bevy_tweening/issues/43) decide if this is acceptable behavior.
```
djeedai, line 946: Left a comment on #43, I'd like to understand what's going on here because it looks like a bug.

src/tweenable.rs
```diff
@@ -894,264 +917,399 @@ mod tests {
...
+                Entity::from_raw(0),
+                &mut event_writer.get_mut(&mut world),
+            ),
+            TweenState::Active
+        );
+        assert!(transform.translation.abs_diff_eq(Vec3::ZERO, 1e-5));
+        assert_eq!(tween.times_completed(), 0);
+    }
+
+    struct ExpectedValues<
```
djeedai, line 969: Ah yes agreed that's a bit ugly, but probably worth it for the understanding of the tests themselves. Thanks!

src/tweenable.rs
```diff
@@ -894,264 +917,399 @@ mod tests {
...
-                tween.clear_completed();
-                assert!(tween.on_completed.is_none());
+            {
+                let just_completed = prev_times_completed != times_completed;
+                prev_times_completed = times_completed;
+
+                let mut event_reader = event_reader.get_mut(&mut world);
+                if just_completed {
+                    assert!(!event_reader.is_empty());
+                    for event in event_reader.iter() {
```
djeedai, line 1094: Should we assert there's a single event here? Since the docs got changed to state that.

src/tweenable.rs
```diff
@@ -894,264 +917,399 @@ mod tests {
...
+            },
+        )
+        .with_direction(TweeningDirection::Forward)
+        .with_repeat_count(RepeatCount::Finite(3))
+        .with_repeat_strategy(RepeatStrategy::Repeat);
+
+        let expected_values = ExpectedValues {
+            deltas: once(Duration::ZERO).chain(repeat(Duration::from_secs_f32(1. / 3.)).take(10)),
+            progress: successors(Some(0.), |progress| Some(f32::min(3., progress + 1. / 3.))),
+            // TODO this is totally wrong due to float precision, should be fixed in upcoming
```
djeedai, line 1162: How do you want to address that? Can we just avoid the edge case by changing the deltas? I understand that 3 steps of 1/3 second might or might not complete the first 1-second loop, but that's just how floating point arithmetic works, we can't really make any strong guarantee for that in the API, can we?

djeedai: Like, if we do `delta = 0.32` then we should avoid any floating-point error over 10+ loops, so would have a deterministic test.

src/tweenable.rs
```diff
@@ -894,264 +917,399 @@ mod tests {
...
+                end: Vec3::ONE,
+            },
+        )
+        .with_direction(TweeningDirection::Forward)
+        .with_repeat_count(RepeatCount::Infinite)
+        .with_repeat_strategy(RepeatStrategy::Repeat);
+
+        let expected_values = ExpectedValues {
+            deltas: [duration * 10, Duration::MAX, Duration::MAX],
+            progress: once(10.).chain(repeat(completions as f32)),
```
djeedai, line 1195: Note we have the same issue here, since `duration`  is 4/3 so irrational, we have no strong guarantee that 10 iterations land _exactly_ on the completed state (short of diving into IEEE and proving it for that particular case, but I want to avoid that).

src/tweenable.rs
djeedai: Can we write the test as if there was no bug, and merge a failing test, rather than merge a test that ensure there is a bug? I'm uncomfortable having tests wrong on purpose.

djeedai: Or, comment out the test for now if the concern is CI will fail.

djeedai (commented):
Ok I was not convinced at first, but that does look a lot easier to read now indeed. I've left a couple of comments, mostly nit-picking that you may skip if you wish, but I'd like to clarify #43 and the negative transform thing first before merging:
- on #43 it feels like there's a bug maybe?
- on the negative transform thing, I'd rather comment out a correct but failing test, than merge a test passing but asserting a wrong result.

src/tweenable.rs
SUPERCILEX: Replied: https://github.com/djeedai/bevy_tweening/issues/43#issuecomment-1210685361. This isn't related to anything we've been doing, but the tests are now getting better and can spot this kind of thing. :)

src/tweenable.rs
SUPERCILEX: Yup, agreed!

src/tweenable.rs
SUPERCILEX: Good point, done.

src/tweenable.rs
SUPERCILEX: Well in theory this should work because we're using durations directly, so I'm hoping #38 will fix it. If not, I'll investigate what's going on and we can probably just use 0.32 like you said.

SUPERCILEX: Oh wait no I see what you mean, the `1. / 3.` is what's hosing us. Will tweak it a little bit.

src/tweenable.rs
SUPERCILEX: Fixed.

src/tweenable.rs
SUPERCILEX: I actually disagree on this one, opinions forthcoming.

Tests validate behavior, not correctness — only a human can assert correctness. In other words, tests document the current behavior.

By removing the tests, we lose the ability to see how the code currently works. That has value in and of itself: someone believing something is broken can find a test that confirms their understanding of the current behavior and then propose new behavior that the fix must adhere.

Furthermore, if there are no existing tests, when a change comes to "fix" the behavior we won't have old tests to validate the bug was actually fixed. That is, adding a new tests prevents you from seeing the transition from incorrect to correct, meaning you can't verify that the test is doing anything at all because you as the reviewer have never seen it fail. Again, this transition provides effortless documentation of the change in behavior that would otherwise be invisible... are the new tests just adding more coverage or validating some new behavior? Who knows!

Adding tests that validate behavior, right or wrong, document the state of the world and its transitions.

SUPERCILEX:
Ok I've been thinking about these tests some more and I'm realizing that I've been dumb. What I'm trying to do is replicate expects tests but I'm forcing us to write code to generate the goldens instead of just using a golden file test. I'm going to switch to that which should be way less code and even easier to understand.

SUPERCILEX:
@djeedai Ok, WDYT?

djeedai:
> @djeedai Ok, WDYT?

I think I don't like goldens conceptually. I can see the value for something like a codegen tool, or some logs, or other text-heavy things. But using that to compare numerical values output as text sounds pretty contrived when you can compare the values directly. This also removes the value of having an `assert!()` telling you exactly _where_ the test failed; I guess you can find it back diff'ing the stdout, but _e.g._ CI probably won't tell you right away. `rstest` was fine, but personally I don't like that new iteration.

SUPERCILEX:
> This also removes the value of having an assert!() telling you exactly where the test failed

I *strongly* disagree here. The asserts are basically useless which is part of why I'm trying to get rid of them. Let me demonstrate with an incorrect completion count.

Here's what a failing golden test looks like:

```
failures:

---- tweenable::tests::tween_tick_loop_finite stdout ----

Goldenfile diff for "tween_tick_loop_finite.stdout":
To regenerate the goldenfile, run
    REGENERATE_GOLDENFILES=1 cargo test
------------------------------------------------------------
thread 'tweenable::tests::tween_tick_loop_finite' panicked at 'assertion failed: `(left == right)`'
...

Differences (-left|+right):
 1.333333332s/1s elapsed
 Direction: Forward
 State: Active
 Progress: 1.3333334
-Total completions: 0
+Total completions: 1
 Transform: Transform { translation: Vec3(1.3333334, 1.3333334, 1.3333334), rotation: Quat(0.0, 0.0, 0.0, 1.0), scale: Vec3(1.0, 1.0, 1.0) }
 Event received: TweenCompleted { entity: 42v0, user_data: 69420 }
...
```

The OG tests:

```
failures:

---- tweenable::tests::tween_tick stdout ----
TweeningType: count=Finite(1) strategy=Repeat dir=Forward
Expected: progress=0.2 factor=0.2 times_completed=0 direction=Forward state=Active just_completed=false translation=Vec3(0.2, 0.2, 0.2)
Expected: progress=0.4 factor=0.4 times_completed=0 direction=Forward state=Active just_completed=false translation=Vec3(0.4, 0.4, 0.4)
Expected: progress=0.6 factor=0.6 times_completed=0 direction=Forward state=Active just_completed=false translation=Vec3(0.6, 0.6, 0.6)
Expected: progress=0.8 factor=0.8 times_completed=0 direction=Forward state=Active just_completed=false translation=Vec3(0.8, 0.8, 0.8)
Expected: progress=1 factor=1 times_completed=0 direction=Forward state=Completed just_completed=true translation=Vec3(1.0, 1.0, 1.0)
thread 'tweenable::tests::tween_tick' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', src/tweenable.rs:1096:21
```

The previous iteration of the new framework:

```
failures:

---- tweenable::tests::tween_tick_loop_finite stdout ----
thread 'tweenable::tests::tween_tick_loop_finite' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`: times_completed', src/tweenable.rs:1058:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

For me, the goldens are hands down the easiest to understand. The OG one requires you to look at the code to find out which value is wrong while the previous iteration clearly needs some context on the current state of the tween. That aside, I think this goes to your next point:

> But using that to compare numerical values output as text sounds pretty contrived when you can compare the values directly.

I kind of agree here. But! The problem I'm trying to solve is understandability and ease of test addition. The OG tests are pretty much impossible to understand with all the indexing and branching (which also means it's impossible to add/update the tests). In the previous iteration, I think it became reasonable to understand which values were expected, but it was still hard to reason about the state context (i.e. after the 7th tick, what should my completion count be?). That meant writing the tests was still a pain because you have to come up with the math to generate the right values and it's super easy to make off-by-one errors or just goof the math. I do agree that the goldens are a bit of a weird approach, but they make writing tests effortless (just create a tween and say what your deltas are) and most importantly they're super easy to audit (they read like a story).

I generally bias towards doing whatever makes the tests easiest to write, because that will lead to the most coverage and therefore the most bugs found.

---

Alrighty, WDYT now? :grin: Any of that convincing?

djeedai:
I like the idea and rationale to push for tests which are both easier to understand and to maintain. But I'm not convinced we need goldens for that.

I think the value argued here is that the version with goldens outputs a human understandable text output that developers can look at to get some context on the failure. On the other hand, the fact goldens use text compare to check correctness is not required, and probably not wanted at all. Conversely, the assert version misses clarity for developers.

So can we have the best of both worlds? Can we output to console the same paragraph of summary text information for each step, and then right after that use `assert!()` to check the validity of each item? That way we retain the debugging / maintainability of tests and avoid the text comparison. I think if we output anything with `println!()`, the test will by default swallow the output if it passes, and only show the test output if it fails (if I remember cargo correctly), which is exactly what we want I think. Any thoughts on this?

SUPERCILEX:
> I think the value argued here is that the version with goldens outputs a human understandable text output that developers can look at **to get some context on the failure**.

Yes except for the last part. I've been thinking about this off and on and I've discovered the following underlying assumption motivating the golden approach: I believe writing the math to generate correct values is too difficult. That is, generating the sequence of correct values takes too long to think about (so fewer tests will get written: to put numbers on this, writing the 6 tests in [`dffbaec` (#53)](https://github.com/djeedai/bevy_tweening/pull/53/commits/dffbaecaa3ca9647d97e22aa43c71a24234e8ebb) took ~1h30m) and involves too much cognitive overhead (meaning it is too difficult for someone to reason about the correctness of the test). And then the sub-assumptions there are that tests should be easy to write and easy to check for correctness. A failing assert is irrelevant in this scenario because I'm saying you don't know *whether or not it should* have failed. A passing test that's wrong is much scarier than a failing test. (See the appendix for code examples.)

Hence, the golden approach embraces the idea that figuring out what the correct values are is too difficult and just says, "Hey, check out what this program did!" Correctness is then determined by deciding whether or not you agree with what the program *did* rather than pre-determining what the program *should do*.

> On the other hand, the fact goldens use text compare to check correctness is not required, and probably not wanted at all.

I think this misses the key difference presented above: do you want to write a test that checks what the program did or one that determines what the program should do? The argument is that "did" is much easier to write and understand (for correctness) than "should do."

> So can we have the best of both worlds? [...] Any thoughts on this?

If the golden approach is abandoned, I'll definitely do this since it will make understanding the source of failures much easier, but it won't address the difficulty in writing the test or reasoning about its correctness when it's passing.

---

### Appendix

Is the transform correct?

```rust
let tween = Tween::new(
    EaseMethod::Linear,
    Duration::from_secs(1),
    TransformPositionLens {
        start: Vec3::ZERO,
        end: Vec3::ONE,
    },
)
.with_direction(TweeningDirection::Forward)
.with_repeat_count(RepeatCount::Finite(3))
.with_repeat_strategy(RepeatStrategy::Repeat);

let expected_values = ExpectedValues {
    ...
    transforms: successors(Some(0.), |progress| Some(f32::min(3., progress + 1. / 3.)))
        .map(|progress| Transform::from_translation(Vec3::splat(progress))),
};
```

We both know it's not because of #42, but I would argue that's not immediately obvious. You could make it easier to see by expanding out the `successors` call (i.e. `[0., 1./3., 2./3., ...]`), but writing that test would be a *massive* PitA. Furthermore, it doesn't protect you from `[0, 0, 0, 0, 1, 1, 1, 2, 2, 2, 3]` where you have no idea if the numbers start in the right places. Maybe it should be `[0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 3]`, who knows. The golden shows you those values *in context* so you can easily reason about their correctness:

```
Initial tween state:
Lens=TransformPositionLens {
    start: Vec3(
        0.0,
        0.0,
        0.0,
    ),
    end: Vec3(
        1.0,
        1.0,
        1.0,
    ),
}

...

Tick by 333.333333ms:
Completion received for 42v0
1.333333332s/1s elapsed
Direction: Forward
State: Active
Progress: 1.3333334
Total completions: 1
Transform: Transform { translation: Vec3(1.3333334, 1.3333334, 1.3333334), rotation: Quat(0.0, 0.0, 0.0, 1.0), scale: Vec3(1.0, 1.0, 1.0) }
Event received: TweenCompleted { entity: 42v0, user_data: 69420 }
```

djeedai:
> That is, generating the sequence of correct values takes too long to think about (so fewer tests will get written: to put numbers on this, writing the 6 tests in [dffbaec (#53)](https://github.com/djeedai/bevy_tweening/pull/53/commits/dffbaecaa3ca9647d97e22aa43c71a24234e8ebb) took ~1h30m) and involves too much cognitive overhead (meaning it is too difficult for someone to reason about the correctness of the test). And then the sub-assumptions there are that tests should be easy to write and easy to check for correctness.
> [...]
> Hence, the golden approach embraces the idea that figuring out what the correct values are is too difficult and just says, "Hey, check out what this program did!" Correctness is then determined by deciding whether or not you agree with what the program did rather than pre-determining what the program should do.

Fair enough, it's hard to write animation tests because there's a fair number of moving variables, and we're testing them in groups so we need to reason about all of them at once. And maybe that's the issue, we should test individual quantities (repeat count, duration, progress, _etc._) in separate tests. But that doesn't mean making it easier to write those test should take priority over having tests that are correct.

- Text comparison for floating point values, or for any numeric value for that matters, is never guaranteed to produce anything useful by any library of any language I know (except maybe libraries explicitly trying to provide that edge-case feature, if any). So testing that floating-point values output by the library are correct by looking at their text representation is extremely brittle; there's no telling what factors are influencing the conversion, and certainly Rust doesn't guarantee it, even less so across compiler versions. I do not want this library to rely on such an approach; with the moving of `progress` as a second-class citizen and the use of `Duration` everywhere we were actually moving in the right direction of correctness, where we can hope to achieve exact bit-level accuracy on playback with the mess of floating-point inaccuracies. Here with goldens this would move us backward in the complete opposite direction.
- You said it, and we obviously have a different conclusion on whether this is wanted, but goldens mean that instead of writing a test about what you think the program does and having the test confirm that (and if not, you find a bug and fix it), now you actually tell the test "tell me what the program actually does", and then I'll have a look at a text file and try to figure out if that's correct, and give a single binary yes/no answer for the entire content of a large file containing hundreds of moving data points. If anything, that sounds exactly what you're describing you want to avoid, only just worst. I simply cannot see how this can ever prevent wrong assumptions and writing incorrect tests. It's already hard enough to write an unbiased test, but now if a developer needs to analyze a 50- to 100-line file instead of focusing on the value a single `assert!()` should have they're bound to make an error and assume the golden is correct when it's not.

> I think this misses the key difference presented above: do you want to write a test that checks what the program did or one that determines what the program should do? The argument is that "did" is much easier to write and understand (for correctness) than "should do."

Yes, "did" is much easier to write. It's also not what a test should do. I know you disagree with that, but I'll repeat my position which is that a test is here to confirm the developer's assumptions of how the code they wrote is supposed to behave. You have a mental model, convert it into some code, then write a test that says "here I wrote the code to achieve this result". And then you run the test and confirm that assumption. And to the extreme, with test-driven development, you actually write the test first.

I see _very_ little value in testing what the program actually does (as opposed to what it was meant to do):
- if it's doing what's expected, then it's actually the same as testing the expected behavior as above, so there's nothing to discuss.
- if not, then you have either of:
  - the only real case of interest, an unexpected behavior, and here "unexpected" can only means "differs from documentation" because that's the only other source of behavior documenting that is shared by everyone, developers and users.
  - a documented behavior, and then that's not a bug but a feature so there's nothing to discuss.

In short, testing the actual behavior serves as a second set of documentation (you said it yourself) that necessarily conflicts with the textual docs (comments), otherwise you'd have a feature. It's also all too easy to forget about such a "wrong behavior" test in the middle of all the other tests. It encourages users looking at the test code to think of bugs as features and take a dependency on them, backing developers in a corner where the "bug" stops being a bug and can never be fixed again (looking at you Windows API, and so many others). Finally, I don't see where you draw the line between an "acceptable" wrong behavior and an "unacceptable" one. Is the behavior acceptable for users if the library produces "3.999999" instead of the "4.0" that the developer intended? Probably, for most users at least, even if not 100% accurate? What about negative `progress()`? That starts to be quite awkward, most users will not expect this, since this represents a percentage, so they probably won't handle negative values and the library will cause errors in their code. And where do we draw a line here? Can we format the user's HDD when playing an animation and say "a yes, you lost everything, but there's a test that validates that behavior so you should have expected it!"? It's obviously exaggerated, but to illustrate the point that there's no more rule on what the test validating.

With intended testing, you test what the behavior should be, as imagined by whoever wrote the code. If something is found wrong in practice, you either fix it now, or if not possible you log a bug and comment out the test `assert!()` so the rest of the test(s) continue to check the other behaviors and produce some value, and you leave the bug reference (link) in comment. Users browsing the test code see that there's a bug, both because the assert is commented out and because there's a link to the bug, and they can follow its evolution. They can find another reference to the bug later in the CHANGELOG once it's fixed, and know the behavior has been restored.

Also, small digression, but on that topic we should have a clean `main` branch with only expected behavior. The "fixing in next PR" that led to #42 is a surefire way to end up with a bug at release time because either whoever was going to fix it forgot, or whoever releases do so "too fast" without knowing the fix was due.

PR #38, opened 2022-08-03
[WIP] Replace progress APIs in favor of elapsed APIs

SUPERCILEX:
Fixes #42, fixes #31.

PR #39, opened 2022-08-03
Carry time delta between tweens in a Sequence

### LINKED ISSUE #36: `Sequence::tick` accumulates timing error whenever a tween completes
https://github.com/djeedai/bevy_tweening/blob/fd86ec51235f1065960de3ed49684c0f2149a259/src/tweenable.rs#L672

`Sequence::tick` spends its entire `delta` time on its current tween with no carry-over. If for example `delta` is 16ms and there are only 2ms left to complete the current tween, then the current tween will get ticked for the entire 16ms and the next tween will start from zero instead of starting 14ms in. Every time the `Sequence` crosses a tween boundary, the error accumulates and the sequence drifts further and further from where it should be.

Gyrobifastigium:
Fixes #36.

### DIFF
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -1243,6 +1243,45 @@ mod tests {
         }
     }
 
+    /// Test crossing tween boundaries in one tick.
+    #[test]
+    fn seq_tick_boundaries() {
+        let mut seq = Sequence::new((0..3).map(|i| {
+            Tween::new(
+                EaseMethod::Linear,
+                TweeningType::Once,
+                Duration::from_secs(1),
+                TransformPositionLens {
+                    start: Vec3::splat(i as f32),
+                    end: Vec3::splat((i + 1) as f32),
+                },
+            )
+        }));
+        let mut transform = Transform::default();
+
+        // Dummy world and event writer
+        let mut world = World::new();
+        world.insert_resource(Events::<TweenCompleted>::default());
+        let mut system_state: SystemState<EventWriter<TweenCompleted>> =
+            SystemState::new(&mut world);
+        let mut event_writer = system_state.get_mut(&mut world);
+
+        // Tick halfway through the first tween, then in one tick:
+        // - Finish the first tween
+        // - Start and finish the second tween
+        // - Start the third tween
+        for delta in [0.5, 2.0] {
+            seq.tick(
+                Duration::from_secs_f32(delta),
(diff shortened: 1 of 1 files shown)


src/tweenable.rs
```diff
@@ -660,28 +660,26 @@ impl<T> Tweenable<T> for Sequence<T> {
...
             let tween = &mut self.tweens[self.index];
-            let tween_state = tween.tick(delta, target, entity, event_writer);
-            if tween_state == TweenState::Completed {
-                tween.rewind();
-                self.index += 1;
-                if self.index >= self.tweens.len() {
-                    state = TweenState::Completed;
-                    self.times_completed = 1;
-                }
+            let tween_remaining = tween.duration().mul_f32(1.0 - tween.progress());
```
djeedai, line 671: Can we avoid using `progress` and multiplications, and just use arithmetic on `Duration`s? We're trying to make progress a second-class concept. That's also probably slower, and has more chance to accumulate errors.

djeedai (changes requested):
Thanks for the contribution! That looks ok but I'd rather have test to make sure this works as expected, and doesn't regress in the future. Could you please add something, even very simple?

src/tweenable.rs
Gyrobifastigium: Technically calculating `tween_remaining` is only needed for the first tween in the loop because the next tweens should all have `tween.progress() == 0`, so I could move it out of the loop.

Other options to calculate `tween_remaining`:

1. Subtract the `Duration`s of all the previous tweens from `self.time`.
2. Track the current tween's progress in a new field.

Option 1 avoids `progress()` but takes linear time. Option 2 avoids `progress()` and takes constant time, but makes `Sequence` a little more complex + uses some memory. Option 2 seems better to me.

djeedai: Ah sorry yes it's because we're missing #38, which I think is essentially option 2. I need to find some time to go review and merge those PRs so we have a better base to iterate on (and before we upgrade to 0.8).

djeedai: Let's leave the progress for now.

Gyrobifastigium:
> Thanks for the contribution! That looks ok but I'd rather have test to make sure this works as expected, and doesn't regress in the future. Could you please add something, even very simple?

Added a test.

djeedai (approved):

PR #45, opened 2022-08-04
Upgrade to Bevy 0.8

djeedai:
Upgrade to the latest released version 0.8.0 of Bevy.

Add a new `bevy_asset` feature to enable animation of Bevy assets
(types implementing the `Asset` trait). Bevy 0.8 does not contain
`bevy_asset` in its defaul features, so this new feature reflects that
new possibility to build Bevy and 🍃 Bevy Tweening without support
for the `bevy_asset` crate. The new feature is enabled by default for
discoverability and to prevent a behavior breaking change.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,6 +56,10 @@ jobs:
         run: cargo test --no-default-features --features="bevy_sprite"
         env:
           CARGO_INCREMENTAL: 1
+      - name: Build & run tests (asset)
+        run: cargo test --no-default-features --features="bevy_asset"
+        env:
+          CARGO_INCREMENTAL: 1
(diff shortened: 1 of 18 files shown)


djeedai:
@SUPERCILEX this change reverts all commits from yesterday, so we get back to a clean and tested base. I choose this option to keep the history linear and keep the release tag on the `main` branch; making a new branch would have created a release outside the `main` branch, which makes things awkward. I'll re-apply those 3 commits once the `0.5.0` release is out, which I plan to do today or tomorrow after some last checks that everything is in order. Let me know if you have any comment on this plan, otherwise I'll move forward. Thanks!

SUPERCILEX (approved):
I'm onboard with this plan!

PR #59, opened 2022-09-30
add an example to show menu hover animation

shuoli84:


djeedai (changes requested):
Thanks for your contribution! I think it would be better to merge this with the existing menu example; that would make one less example to maintain, and make the existing example more full-featured. Could you please merge your code into the existing menu example? Thanks!

shuoli84:
@djeedai sure, just merged, also add a new system in order to ensure appearance animation not overstepped with interaction animation.

examples/menu.rs
```diff
@@ -3,6 +3,22 @@ use bevy_inspector_egui::WorldInspectorPlugin;
...
+/// The menu in this example has two set of animations:
+/// one for appearance, one for interaction. And interaction animation
+/// should only enabled after appearance animation finished.
+/// The logic is handled as:
+/// 1. Appearance animation send a TweenComplete event with INIT_TRANSITION_DONE
+/// 2. in system `enable_interaction_after_initial_animation`, it add a label component
+/// `InitTransitionDone` to button component.
+/// 3. `interaction` system only queries button with `InitTransitionDone`
+///
+/// If your menu doesn't has appearance animation, then you can omit step 1 and 2.
```
djeedai, line 21:
```suggestion
/// The menu in this example has two set of animations:
/// one for appearance, one for interaction. Interaction animations
/// are only enabled after appearance animations finished.
///
/// The logic is handled as:
/// 1. Appearance animations send a `TweenComplete` event with `INIT_TRANSITION_DONE`
/// 2. The `enable_interaction_after_initial_animation` system adds a label component
/// `InitTransitionDone` to any button component which completed its appearance animation,
/// to mark it as active.
/// 3. The `interaction` system only queries buttons with a `InitTransitionDone` marker.
```

examples/menu.rs
```diff
@@ -102,3 +129,82 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
+                    ButtonLabel::Quit => {
+                        println!("Quit clicked");
+                    }
+                }
+            }
+            Interaction::Hovered => {
+                *color = HOVER_COLOR.into();
+                commands.entity(entity).insert(Animator::new(Tween::new(
+                    EaseFunction::QuadraticIn,
+                    Duration::from_secs_f32(0.2),
```
djeedai, line 188: Prefer `Duration::from_millis(200)` which doesn't have rounding errors. I'm in the process of converting all examples and the internal library code to avoid floating point calculations as much as possible, for accuracy reasons.

djeedai (changes requested):
Thanks for the merge! That looks really much better.

There's a wrong usage pattern of adding an `Animator` again and again, that I've seen from other users too, so I'd rather we don't propagate its use in examples. Other than that just minor things but looks good overall, thanks!

examples/menu.rs
```diff
@@ -102,3 +129,82 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
+    mut commands: Commands,
+    mut reader: EventReader<TweenCompleted>,
+) {
+    for event in reader.iter() {
+        if event.user_data == INIT_TRANSITION_DONE {
+            commands.entity(event.entity).insert(InitTransitionDone);
+        }
+    }
+
+    reader.clear();
```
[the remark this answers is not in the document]
djeedai: Yes it does, see documentation : https://docs.rs/bevy/latest/bevy/ecs/event/struct.EventReader.html#method.iter

djeedai (approved):

Issue #31, opened 2022-07-11
Get rid of progress (well keep an ext trait with it around) in favor of elapsed


SUPERCILEX:
https://github.com/djeedai/bevy_tweening/pull/19#discussion_r912525741

PR #74, opened 2022-11-10
Added total duration and shared implementations

djeedai:
Exposed the total animation duration via the `TotalDuration` enum and the `Tweenable<T>::total_duration() -> TotalDuration` API.

Added a default implementation for some methods of `Tweenable<T>`, namely `set_progress()`, `progress()`, and `times_completed()`. These shared implementations are used by all built-in tweenables.

Issue: #31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `Targetable`, `ComponentTarget`, and `AssetTarget`, which should be considered private even though they appear in the public API. They are a workaround for Bevy 0.8 and will likely be removed in the future once the related Bevy limitation is lifted.
 - Added the missing `Tween::with_completed()` to raise a callback.
 - Added completion event and callback support to `Delay<T>`, similar to what existed for `Tween<T>`.
+- Added `TotalDuration` and a new `Tweenable<T>::total_duration()` method to retrieve the total duration of the animation including looping.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


djeedai:
@SUPERCILEX is you're around and you have time I'd like your opinion on this change. It partly addresses the idea from #31 of making `set_progress()`, `progress()`, and `times_completed()`, second class citizens. However it does so by writing a default implementation in the `Tweenable<T>` trait directly, rather than using extension traits. The reasons are not only that I find that quite clean _a posteriori_, but also that I tried the extension traits approach and had a few issues:
- It's tricky with generic traits like `Tweenable<T>` as you cannot simply provide a catch-all implementation of the extension trait for all `Tweenable<T>` (https://stackoverflow.com/questions/65622443/implement-a-trait-for-a-generic-trait). You can work around with a macro implementation for all concrete types one by one, but that's not great.
- Because `Animator::tweenable()` returns `&dyn Tweenable<T>` (return type needs to be sized), and similarly `tweenable_mut()` returns `&mut Twenable<T>`, and you'd like to be able to use the extension trait directly on those, _e.g._:
  ```rust
  let progress = animator.tweenable().progress();
  ```
  ...then I don't know of any valid way to provide an implementation of the trait that makes that happen. You'd ideally implement the extension trait for those 2 types `&dyn Tweenable<T>` and `&mut dyn Tweenable<T>` but it becomes at best hard and probably nonsensical as you end up having _e.g._:
  ```rust
  impl ExtensionTrait for &dyn Tweenable<T> {
    fn progress(&self) -> f32 {} // &self == &&dyn Tweenable<T>
    fn set_progress(&mut self) {} // &mut self == &mut &dyn Tweenable<T>, and cannot mutate the tweenable itself
  }
  ```
  So then you can split into a non-mut trait and a mut trait but that becomes ugly. And again, I'm not sure that even makes sense. Maybe there's another way, but that look all convoluted and error-prone.

PR #89, opened 2023-06-13
Fixed a panic in text color lens

wainwrightmark:
Hi. The `TextColorLens` can panic if the text section it's looking for isn't there. I just got it to do nothing in that case instead.


wainwrightmark:
> Thanks! Can you please add a line in the CHANGELOG for this?

Have done. Also updated the tests as coverage had gone done.

djeedai (approved):

PR #90, opened 2023-06-24
Updated to bevy 0.10.1

LemonjamesD:
Update the crate and bevy version to use bevy 0.10.1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy_tweening"
-version = "0.7.0"
+version = "0.7.1"
 authors = ["François Mockers <mockersf@gmail.com>", "Jerome Humbert <djeedai@gmail.com>"]
 edition = "2021"
 description = "Tweening animation plugin for the Bevy game engine"
@@ -25,7 +25,7 @@ bevy_text = ["bevy/bevy_text", "bevy/bevy_render", "bevy/bevy_sprite"]
 
 [dependencies]
 interpolation = "0.2"
-bevy = { version = "0.10", default-features = false }
+bevy = { version = "0.10.1", default-features = false }
 
 [dev-dependencies]
 bevy-inspector-egui = "0.18"


djeedai (commented):
There's no reason to do that. In fact, this only prevents users on 0.10.0 from using the crate, for no good reason.

LemonjamesD:
> There's no reason to do that. In fact, this only prevents users on 0.10.0 from using the crate, for no good reason.

So using minor releases of bevy is discouraged? Any particular version of this crate can be argued "it prevents users of x.x.x from using the crate." It seems very anti-update to me for no reason. 

Does it really hurt people to use a minor change of bevy? It took me 3 seconds to fork and change versions.

djeedai:
> So using minor releases of bevy is discouraged?

Nobody said that. I don't know how you came to that conclusion, but it feels like maybe you're misunderstanding how cargo dependency versioning works. `bevy = "0.10"` means any version >= 0.10.0 and < 0.11.0, so in particular 0.10.1 is already included. The only reason to do the change you propose would be if this crate used a feature only available in 0.10.1 but not in 0.10.0, which to the best of my knowledge is not the case.

> Any particular version of this crate can be argued "it prevents users of x.x.x from using the crate."

No, because the version of this crate is irrelevant, we're talking about a dependency version here.

> It seems very anti-update to me for no reason.

I already explained it twice, this change forces users of this library to upgrade to Bevy 0.10.1 even if they don't want to, and there's absolutely no benefit I can think of in return since Bevy Tweening works with both versions as far as I know. Why would this library force an update and annoy people? You can upgrade your own game to use Bevy 0.10.1 without this change. Or you can keep using 0.10.0, your choice. With your proposed change the library prevents the users from making that choice.

LemonjamesD:
> > So using minor releases of bevy is discouraged?
> 
> Nobody said that. I don't know how you came to that conclusion, but it feels like maybe you're misunderstanding how cargo dependency versioning works. `bevy = "0.10"` means any version >= 0.10.0 and < 0.11.0, so in particular 0.10.1 is already included. The only reason to do the change you propose would be if this crate used a feature only available in 0.10.1 but not in 0.10.0, which to the best of my knowledge is not the case.
> 
> > Any particular version of this crate can be argued "it prevents users of x.x.x from using the crate."
> 
> No, because the version of this crate is irrelevant, we're talking about a dependency version here.
> 
> > It seems very anti-update to me for no reason.
> 
> I already explained it twice, this change forces users of this library to upgrade to Bevy 0.10.1 even if they don't want to, and there's absolutely no benefit I can think of in return since Bevy Tweening works with both versions as far as I know. Why would this library force an update and annoy people? You can upgrade your own game to use Bevy 0.10.1 without this change. Or you can keep using 0.10.0, your choice. With your proposed change the library prevents the users from making that choice.

Ah ha I'm a moron. I thought 0.10 was just an alias for 0.10.0. Sorry for my misunderstanding I thought you just didn't like patches for no reason.

djeedai:
> I thought 0.10 was just an alias for 0.10.0. Sorry for my misunderstanding I thought you just didn't like patches for no reason.

`0.10` _is_ an alias for `0.10.0` and `^0.10`. And they both mean `>=0.10.0, <0.11.0`. This is different from `=0.10.0` which specifies an _exact_ version. See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html.

PR #92, opened 2023-07-10
Upgrade to Bevy 0.11

arendjr:
Title says it all :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy_tweening"
-version = "0.7.0"
+version = "0.8.0"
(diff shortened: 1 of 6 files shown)


djeedai (approved):
Amazing, thanks a lot for that!

arendjr:
I saw in the CI I had still missed some references. I've updated the deps in the benchmarks too now, as well as some doc links. Only it seem `bevy_egui` still needs an update to be compatible with Bevy 0.11. I see they're working on it, but we may need to wait a little until it's released: https://github.com/mvlabat/bevy_egui/commit/22bf74a5489486c95d6f89a0769dd0c1f26edf59

djeedai (approved):

Issue #79, opened 2022-11-22
Sequence with repeat_count


enaut:
Hey,

I was trying to have an object zoom into existence then wiggle infinitely... I tried doing that with:

```rust
Tween::new(
            EaseFunction::QuadraticInOut,
            Duration::from_millis(1000),
            TransformScaleLens {
                start: Vec3::ZERO,
                end: Vec3::ONE,
            },
        )
        .then(
            Tween::new(
                EaseFunction::QuadraticInOut,
                Duration::from_millis(200),
                TransformRotateZLens {
                    start: 0.05,
                    end: -0.05,
                },
            )
            .with_repeat_strategy(bevy_tweening::RepeatStrategy::MirroredRepeat)
            .with_repeat_count(RepeatCount::Infinite),
        )
```

But when Running the first animation works well, however the second one panics with:

```
thread 'Compute Task Pool (13)' panicked at 'overflow when subtracting durations', library/core/src/time.rs:930:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The same happens if `repeat_count` is set to a finite number even when set to `1`.

It might be that this is "known" or "intended" but if so it should probably be in the documentation for the `then` method or the `Sequence` type.

The Behaviour I expected would have been an infinite sequence first the zoom then forever wiggling...

djeedai:
Yes i don't think we support appending a finite and an infinite tweenables, because although that looks fine when playing forward, there's all sorts of undefined behaviors when playing backward or seeking to a particular point. I'm still perplex on the actual error though, looks like a bug.

djeedai:
I mean, we could revert that I guess, but again that doesn't solve the design question of "what do we do when the mirrored tween reaches back its start point?". If we mirror again then we're stuck on that tween, and any other tween in the `Sequence` will never play again. If we don't, then we break the contract of the the mirroring and/or infinite looping. Either way something goes in a potentially unexpected way for the user. I'm open to suggestions here, because I don't see a good alternative.

enaut:
Just a thought wouldn't it make sense to add a `finally(end: Tweenable)` method to the `Sequence`. The parameter to this method is always added after all the others and can be infinite. This would also be great for fade-out effects or similar.
[...]

PR #95, opened 2023-07-11
Infinite mirrored repeats crash

SergioRibera:
I have seen that there was not a PR about it and I think it is something that had to be fixed, it turns out that if we have an infinite or repeated sequence there is a problem when calculating the remaining time, for this a solution was proposed [here](https://github.com/djeedai/bevy_tweening/issues/79#issuecomment-1383284902) and [here](https://github.com/sibsibsib/bevy_tweening/commit/97b5eac01b7472dec813e88c45e00556d646953c) was developed the implementation, apparently they had not opened a PR so here I am to propose the PR updated to the last commit of the main branch in the project

### DIFF
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -839,7 +839,7 @@ impl<T> Tweenable<T> for Sequence<T> {
         self.elapsed = self.elapsed.saturating_add(delta).min(self.duration);
         while self.index < self.tweens.len() {
             let tween = &mut self.tweens[self.index];
-            let tween_remaining = tween.duration() - tween.elapsed();
+            let tween_remaining = tween.duration().mul_f32(1.0 - tween.progress());
             if let TweenState::Active = tween.tick(delta, target, entity, events) {
                 return TweenState::Active;
             }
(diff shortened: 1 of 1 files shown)


djeedai:
Thanks for the contribution. Unfortunately I think this requires more discussion and an agreed plan of action first.

There's a number of issues with the implemented approach that I don't like.

1. The behavior is maybe useful in some cases, but is definitely unexpected. It's a mix between the `Sequence` and `Tracks` behaviors, which depends on top of that on other parameters. I'm afraid this will just confuse most users. Even the OP admits it "weird".

2. As explained in several other places, the issue is not to solve in isolation those individual issues; this is reasonably easy. The real issue is to design something consistent. And all the difficulty comes from other APIs ignored in most discussions. For example, what happens if the user jumps to a specific point in time with `set_elapsed()`? And do we trigger completion events in that case? What about playback in reverse direction?

I appreciate sequences with an infinite loop is a common use case several users asked for. But I have yet to hear about a reasonable design that 1) doesn't introduce unexpected behaviors, and 2) remains consistent across the various APIs available to manipulate the tweenable playback. I'm fine making concessions about having to raise a warning at runtime if the user crafts an unreasonable animation, if we can't find a way to prevent that by design at build time. But at least I want to prevent users from shooting themselves in the foot with an animation there's no clear mental model for; this would be the worst we can achieve here.

As a minor extra issue, the `progress`-based API is kind of deprecated, and shouldn't be used as the source of truth.
 It lacks precision over the `Duration`-based one, due to floating point. It can remain as a utility on top of the `Duration`-based one but we already worked to remove the use of `progress` internally, and this change undoes it.

SergioRibera:
A perfect, indeed they are cases that are not contemplated in the implementation.

Is there a channel where this is being discussed? I would like to see and participate, maybe we can make this solution faster 😁

djeedai:
The Discord channel for this crate is on Bevy's official server: https://discord.com/channels/691052431525675048/1036006196659900497

djeedai:
I'm going to close this one, as I don't think there's consensus on the approach. Sorry about that :) Please feel free to reopen and argument if there's a strong opinion that this is indeed the best approach, and why?

PR #97, opened 2023-07-19
BackgroundColor support

Sjael:
I's pretty common to need the ability to fade UI elements/images on and off, so I added a lens for BackgroundColor. 

Simply tween the alpha value for two Color::rgb().set_a()

```rust
let selected_color = Color::RED;
let tween_opac = Tween::new(
    EaseFunction::QuadraticIn,
    Duration::from_secs(1),
    UiBackgroundColorLens {
        start: selected_color.set_a(0.9),
        end: selected_color.set_a(0.0),
    },
);
let animator = Animator::new(tween_opac);
```
It's very nice for flashing something on screen and removing after a delay, like a killfeed notification.

```rust
let selected_color = Color::RED;
let tween_opac_in = Tween::new(
    EaseFunction::QuadraticIn,
    Duration::from_millis(500),
    UiBackgroundColorLens {
        start: selected_color.set_a(0.0),
        end: selected_color.set_a(0.9),
    },
);
let tween_opac_out = Tween::new(
    EaseFunction::QuadraticIn,
    Duration::from_secs(1),
    UiBackgroundColorLens {
        start: selected_color.set_a(0.9),
        end: selected_color.set_a(0.0),
    },
).with_completed_event(
    TweenEvents::KillNotifEnded as u64,
);
let animator =  Animator::new(tween_opac_in
    .then(Delay::new(
        Duration::from_secs(15),
    ))
    .then(tween_opac_out)
);
```

This would also work for going between any 2 colors, but I think the opacity will be most used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added built-in support for the `BackgroundColor` bevy component, under the `bevy_ui` feature.
+
 ## [0.8.0] - 2023-07-12
 
 ### Changed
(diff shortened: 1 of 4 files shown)


djeedai (approved):
Hi, thanks for the change!

I'm generally against adding more lenses, as really the system is designed for users to write their own, and I don't want to have to maintain every single use case. That being said, here I appreciate how common this is, and is in a sense an unfortunate regression of Bevy newtyping all its UI type a few releases ago. So OK for this change.

djeedai (commented):
Can you please update the CHANGELOG, and ideally add a small test for that new lens? Thanks!

djeedai (approved):

djeedai (changes requested):
Need CHANGELOG update

djeedai (approved):
I added the CHANGELOG change myself just to complete the PR as we're nearing the Bevy 0.12 release and I want to be ready :) Thanks!

PR #98, opened 2023-07-29
Add a tweening settings resource

SecretPocketCat:
This adds a setting to always fire the tween done event (or rather a default user data value) using a settings resource). This would of course be doable by always setting the done event user data, but if that's error-prone if the data is not used, only the event itself.

The resource is initialized in the plugin and can be replaced by inserting the resource after adding the plugin which is the idiomatic bevy approach I think. The resource init could also be skipped and the systems could use it optionally or the settings could be add to the plugin along with a new constructor, but that would be a breaking change.

If this gets merged, I'd also like to add a timescale setting to the settings resource.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -488,9 +488,9 @@ impl<T: Asset> AssetAnimator<T> {
     }
 }
 
-/// Default tweening settings
+/// Tweening settings
 #[derive(Resource, Default, Clone, Copy)]
-pub struct TweenSettings {
+pub struct TweeningSettings {
     /// Default completed event user data
     /// setting this value will always fire the completed event without having to set it explicitly
     pub default_completed_event_data: Option<u64>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -488,6 +488,14 @@ impl<T: Asset> AssetAnimator<T> {
     }
 }
 
+/// Default tweening settings
+#[derive(Resource, Default, Clone, Copy)]
+pub struct TweenSettings {
+    /// Default completed event user data
+    /// setting this value will always fire the completed event without having to set it explicitly
+    pub default_completed_event_data: Option<u64>,
+}
+
 #[cfg(test)]
 mod tests {
     #[cfg(feature = "bevy_asset")]
--- a/src/plugin.rs
+++ b/src/plugin.rs
@@ -4,7 +4,7 @@ use bevy::{ecs::component::Component, prelude::*};
 
 #[cfg(feature = "bevy_asset")]
 use crate::{tweenable::AssetTarget, AssetAnimator};
-use crate::{tweenable::ComponentTarget, Animator, AnimatorState, TweenCompleted};
+use crate::{tweenable::ComponentTarget, Animator, AnimatorState, TweenCompleted, TweenSettings};
 
 /// Plugin to add systems related to tweening of common components and assets.
 ///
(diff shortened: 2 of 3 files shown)


djeedai:
Hi- Thanks for taking the time to write this change.

I'm pretty skeptical about it though. It seems your issue is trying to systematically send a completion event. I'd argue the change doesn't really align with that:

- Setting a default value for the user data defeats the point of having that value in the first place, which is designed to discriminate between the events raised on various components.
- It's very unclear that setting a default value also activates the event sending by default. Yes this is documented, but the semantic of the field doesn't reflect that, as it's focused instead of user data. I'm afraid this will create a lot of confusion.
- There are ways to do this without any change; for example you can simply write a helper function that returns a new `Tween` ans always call `with_completed_event()` on it.
- If the issue is that the event is very useful but requires manual setting, maybe we should consider changing the default instead. Bevy handles event well enough I think that this shouldn't bother too much people not using them hopefully. As far as I understand Bevy itself sends events for plenty of things that users don't actively use by default, but can prove useful in some circumstances, so it wouldn't be unheard of.
- That user data has already been raised by others as awkward and non idiomatic in Rust. I'm not sure I want to further extend its reach.

Overall I don't see any good reason to merge this change, especially given how pervasive it is (seems the settings need to be passed down everywhere). I'd rather discuss a different solution, like mark this as a user issue and do nothing, or change the default value and always send events, or even add a `bool` setting for controlling the default send mode.

SecretPocketCat:
> * Setting a default value for the user data defeats the point of having that value in the first place, which is designed to discriminate between the events raised on various components.
> * It's very unclear that setting a default value also activates the event sending by default. Yes this is documented, but the semantic of the field doesn't reflect that, as it's focused instead on user data. I'm afraid this will create a lot of confusion.

I can't really argue with that, I just wanted to add this as an option without introducing any breaking changes, but the events being fired are based on having some data to actually fire at all.

> * There are ways to do this without any change; for example you can simply write a helper function that returns a new `Tween` ans always call `with_completed_event()` on it.

That seems error-prone and like extra boilerplate to me (if the only thing the fn does is just set that event).
I've always used helper fns for creating tweens, but I've also created tweens in place and I can imagine myself (or someone else if I worked in a team) forgetting to set this and then wondering why is the completion event not firing.

> * If the issue is that the event is very useful but requires manual setting, maybe we should consider changing the default instead. Bevy handles event well enough I think that this shouldn't bother too much people not using them hopefully. As far as I understand Bevy itself sends events for plenty of things that users don't actively use by default, but can prove useful in some circumstances, so it wouldn't be unheard of.

I'd like that, but how would the user data be handled? Would 0 be used as a sentinel value, (which doesn't feel idiomatic/rust-y)?
Does that field get changed to an Option or is dropped entirely? That would be a breaking change.

> Overall I don't see any good reason to merge this change, especially given how pervasive it is (seems the settings need to be passed down everywhere). I'd rather discuss a different solution, like mark this as a user issue and do nothing, or change the default value and always send events, or even add a `bool` setting for controlling the default send mode.

I'd be fine with making the setting a bool, but the question regarding the user data value still applies.

I'd personally prefer making the setting a bool and changing user data to be an option.
Alternatively, it could be a feature, which could avoid the breaking change, but it would be pretty hard to discover that.

SecretPocketCat:
@djeedai Sorry to ping you, but would you mind chiming in once more? If you don't like any of the approaches you proposed, I'm fine with closing this rather then just letting it go stale.

SecretPocketCat:
Closing as stale

Issue #103, opened 2023-10-25
The 'custom component' description is misleading


SecretPocketCat:
As [noted on discord](https://discord.com/channels/691052431525675048/1036006196659900497/1162847997089021973), the the `custom component` part of the readme is a bit misleading and a common source of confusion. The readme should probably be more explicit about the fact that not all built-in components are supported (and probably just link to the plugin source as both an example of adding the system and to show which types are added by default).

djeedai:
Fixed, thanks!

PR #104, opened 2023-10-27
Clarify doc around systems

djeedai:
Clarify the documentation around animation systems, which ones are added automatically by the `TweeningPlugin` and which ones need to be added manually, depending on which feature is active.
- Expand the Setup section of the README to explain the systems need to be added manually for most components and assets, and give a table for built-in ones where the system is automatically added.
- Add a similar section to the rustdoc section of `lib.rs`.

Also fix a couple of doc links.

Bug: #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,35 @@ App::default()
     .run();
 ```
 
+This provides the basic setup for using 🍃 Bevy Tweening. However, additional setup is required depending on the components and assets you want to animate:
+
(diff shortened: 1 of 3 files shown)


djeedai:
@SecretPocketCat if you want to have a look at this change and let me know if the wording around animation systems sounds more clear that way? Thanks

SecretPocketCat:
That's much better, thank you.

I added a few nits in #105 (just missing spaces) and replaced the mention of `built-in Bevy` components in case someone is just skimming the docs, but that's upto you.

PR #108, opened 2023-11-06
Upgrade to Bevy 0.12

arendjr:
I've taken the liberty to bump the crate's own version to 0.9. The compatibility table in the README is also updated.


arendjr:
Just FYI, it seems the failing check is merely a reduction is code coverage, which doesn't seem to be an *actual* reduction, but merely because the test setup itself has become a few lines more compact :)

PR #110, opened 2023-11-09
Add missing docs for `trait Lens` 

tom-ainc:
# Overview

I spent 2+ hours wondering why my implementation of Lens wasn't working for a Bevy built in component (specifically Projection as I was making a zoom in animation), only to eventually find the answer in the README instead of docs.rs.

This PR exists to save others from the same walk around of the internet & is an almost verbatim copy from the aforementioned README.

### DIFF
--- a/src/lens.rs
+++ b/src/lens.rs
@@ -65,6 +65,22 @@ use bevy::prelude::*;
 ///   }
 /// }
 /// ```
+///
+/// /// Any cusotom implementation of `Lens` for animating any component or asset other than those
+/// provided by Bevy Tweening itself: `Transform`, `Sprite`, `ColorMaterial`, `Style`, `Text`,
+/// (either built-in from Bevy itself, or custom) requires manual  scheduling the appropriate
+/// system.
+///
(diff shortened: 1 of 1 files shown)


djeedai:
I'm sorry it took you some time to find this. But I don't think this is the appropriate place to document it. This has nothing to do specifically with this trait but rather with how the library works, and specifically what the `TweeningPlugin` does when added to the `App`. There's already an entire section about this at the package root, which makes it as the homepage on docs.rs. This is also on the README as you noted. I appreciate it's not intuitive, and I'm open to suggestions on how we could make that registration automatic, but at some point we can't document how the library works in every single docs page. I'm fine adding a reference to the docs here if that helps, but I'm not keen on having to maintain yet another variant of the same explanation in yet another location.

https://docs.rs/bevy_tweening/latest/bevy_tweening/#system-setup

#103, #104

djeedai:
I'm going to close this without merging. I think @tom-ainc you Liked my comments and so I assume agreed that it's already documented in more relevant places.

PR #111, opened 2023-11-14
Run a one-shot system upon completion of Tween or Delay

Peepo-Juice:
This PR adds with_completed_system() and other related methods to allow the user to run a one-shot system when an animation or delay completes. 

The main reason why I wanted to add this is to avoid polling for a single non-repeating event; using one-shot systems can be cleaner sometimes. But there are other use-cases like chaining actions or even other animations that cant be chained normally due to limitations in the library. In those cases you may think to use callbacks, but they don't allow you to easily interact with the world, which makes one-shot systems a necessity.

### DIFF
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -1338,7 +1338,7 @@ mod tests {
     use std::sync::{Arc, Mutex};
 
     use bevy::ecs::{
-        component::Tick
+        component::Tick,
         event::Events,
         system::{CommandQueue, SystemState},
     };
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -25,6 +25,7 @@ use crate::{EaseMethod, Lens, RepeatCount, RepeatStrategy, TweeningDirection};
 /// implement [`From`]:
 /// ```no_run
 /// # use std::time::Duration;
+/// # use bevy::ecs::system::Commands;
 /// # use bevy::prelude::{Entity, Events, Mut, Transform};
 /// # use bevy_tweening::{BoxedTweenable, Sequence, Tweenable, TweenCompleted, TweenState, Targetable, TotalDuration};
 /// #
@@ -34,7 +35,7 @@ use crate::{EaseMethod, Lens, RepeatCount, RepeatStrategy, TweeningDirection};
 /// #     fn total_duration(&self) -> TotalDuration  { unimplemented!() }
 /// #     fn set_elapsed(&mut self, elapsed: Duration)  { unimplemented!() }
 /// #     fn elapsed(&self) -> Duration  { unimplemented!() }
-/// #     fn tick<'a>(&mut self, delta: Duration, target: &'a mut dyn Targetable<Transform>, entity: Entity, events: &mut Mut<Events<TweenCompleted>>) -> TweenState  { unimplemented!() }
+/// #     fn tick<'a>(&mut self, delta: Duration, target: &'a mut dyn Targetable<Transform>, entity: Entity, events: &mut Mut<Events<TweenCompleted>>, commands: &mut Commands) -> TweenState  { unimplemented!() }
 /// #     fn rewind(&mut self) { unimplemented!() }
 /// # }
 ///
@@ -590,9 +591,9 @@ impl<T> Tween<T> {
     ///
     /// ```
     /// # use bevy_tweening::{lens::*, *};
-    /// # use bevy::{ecs::event::EventReader, math::Vec3};
+    /// # use bevy::{ecs::event::EventReader, math::Vec3, ecs::world::World, ecs::system::Query, ecs::entity::Entity, ecs::query::With};
     /// # use std::time::Duration;
-    /// let world = World::new();  
+    /// let mut world = World::new();  
     /// let test_system_system_id = world.register_system(test_system);
     /// let tween = Tween::new(
     ///     // [...]
@@ -605,9 +606,9 @@ impl<T> Tween<T> {
     /// )
     /// .with_completed_system(test_system_system_id);
     ///
-    /// fn test_system(query: Query<Entity, With<Enemy>>) {
-    ///    for enemy in query.iter() {
-    ///       println!("Found an enemy!");
+    /// fn test_system(query: Query<Entity>) {
+    ///    for entity in query.iter() {
+    ///       println!("Found an entity!");
     ///    }
     /// }
     /// ```
@@ -720,8 +721,8 @@ impl<T> Tween<T> {
     /// but uses a system registered by [`register_system()`] instead of a callback.
     ///
     /// [`with_completed()`]: Tween::with_completed  
-    /// [`register_system()`]: bevy::ecs::world::World::register_system 
-    pub fn set_completed_system(&mut self, user_data: SystemId) { 
+    /// [`register_system()`]: bevy::ecs::world::World::register_system
+    pub fn set_completed_system(&mut self, user_data: SystemId) {
         self.system_id = Some(user_data);
     }
 
@@ -1163,8 +1164,9 @@ impl<T> Delay<T> {
     ///
     /// ```
     /// # use bevy_tweening::{lens::*, *};
-    /// # use bevy::{ecs::event::EventReader, math::Vec3};
+    /// # use bevy::{ecs::event::EventReader, math::Vec3, ecs::world::World, ecs::system::Query, ecs::entity::Entity};
     /// # use std::time::Duration;
+    /// let mut world = World::new();
     /// let test_system_system_id = world.register_system(test_system);
     /// let tween = Tween::new(
     ///     // [...]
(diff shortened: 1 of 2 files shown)


Peepo-Juice:
I tried my best to include some unit tests to maintain symmetry with the other "with_completed" methods, but I ran into a snag when dealing with the tick tests. I had to create some dummy values in `manual_tick_component` in tweenable.rs to allow it to compile, and currently there's no testing of the execution of the one-shot system itself. I could have made a test for that myself if I ignore the `manual_tick_component()`, but I realized that #37 exists so maybe it can be taken care of in that PR? 

Oh, and I just realized I forgot to update the examples. I can get to that as soon as I get some feedback on what I managed to to commit so far :)

src/tweenable.rs
```diff
@@ -538,6 +541,44 @@ impl<T> Tween<T> {
...
+    /// animation completes. This is similar to the [`with_completed()`],
+    /// but uses a system registered by [`register_system()`] instead of a callback.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use bevy_tweening::{lens::*, *};
+    /// # use bevy::{ecs::event::EventReader, math::Vec3};
+    /// # use std::time::Duration;
+    /// let test_system_system_id = world.register_system(test_system);
```
djeedai, line 556: I think the example won't build because `world` is undefined.

src/tweenable.rs
```diff
@@ -631,6 +672,28 @@ impl<T> Tween<T> {
     pub fn clear_completed_event(&mut self) {
         self.event_data = None;
     }
+
+    /// Enable running a one-shot system upon completion.
+    ///
+    /// If enabled, the tween will run a system via a provided [`SystemId`] when the
+    /// animation completes. This is similar to the [`with_completed()`],
+    /// but uses a system registered by [`register_system()`] instead of a callback.
```
djeedai, line 680: What's `register_system()` here?

djeedai (approved):
I think the change makes sense, thanks a lot! Couple of comments on docs, and I don't know why it seems CI didn't start? I'll check that too as I expect docs won't build. Other than that looks good to me.

src/tweenable.rs
Peepo-Juice: Sorry for the delayed response, been dealing with personal crap.

That is [bevy::ecs::world::World::register_system](https://docs.rs/bevy/latest/bevy/ecs/world/struct.World.html#method.register_system).

src/tweenable.rs
Peepo-Juice: Tbh I had no idea what I was doing here, I thought it was just a harmless comment. I feel like I'm being ignorant of something there. I suppose we can just add:

```
let world = World::new();
```

Peepo-Juice (commented):
This was as a result of me trying to update my local repo and manage a conflict. I feel like I screwed up and created this ugly commit. My bad! I'm still getting the hang of Git.

src/tweenable.rs
```diff
@@ -535,6 +538,44 @@ impl<T> Tween<T> {
...
+    /// fn test_system(query: Query<Entity, With<Enemy>>) {
+    ///    for enemy in query.iter() {
+    ///       println!("Found an enemy!");
+    ///    }
+    /// }
+    /// ```
+    ///
+    /// [`with_completed_system()`]: Tween::with_completed_system
+    #[must_use]
+    pub fn with_completed_system(mut self, user_data: SystemId) -> Self {
```
djeedai, line 574:
```suggestion
    pub fn with_completed_system(mut self, system_id: SystemId) -> Self {
```

src/tweenable.rs
```diff
@@ -456,6 +458,7 @@ impl<T> Tween<T> {
             lens: Box::new(lens),
             on_completed: None,
             event_data: None,
+            system_data: None,
```
djeedai, line 461:
```suggestion
            system_id: None,
```

src/tweenable.rs
djeedai: I meant, if you don't tell rustdoc what `register_system()` is, the link will fail to build because rustdoc will not guess that it's a function on Bevy's `World`.

src/tweenable.rs
```diff
@@ -628,6 +669,28 @@ impl<T> Tween<T> {
...
+    /// Enable running a one-shot system upon completion.
+    ///
+    /// If enabled, the tween will run a system via a provided [`SystemId`] when the
+    /// animation completes. This is similar to the [`with_completed()`],
+    /// but uses a system registered by [`register_system()`] instead of a callback.
+    ///
+    /// See [`with_completed_system()`] for details.
+    ///
+    /// [`set_completed_system()`]: Tween::set_completed_system
+    pub fn set_completed_system(&mut self, user_data: SystemId) {
```
djeedai, line 682:
```suggestion
    /// [`set_completed_system()`]: Tween::set_completed_system
    /// [`register_system()`]: bevy::ecs::world::World::register_system
    pub fn set_completed_system(&mut self, user_data: SystemId) {
```

src/tweenable.rs
```diff
@@ -628,6 +669,28 @@ impl<T> Tween<T> {
...
+        self.system_data = Some(user_data);
+    }
+
+    /// Clear the system that will execute when the animation completes.
+    ///
+    /// See also [`set_completed_system()`].
+    ///
+    /// [`clear_completed_system()`]: Tween::clear_completed_system
```
djeedai, line 690:
```suggestion
    /// [`set_completed_system()`]: Tween::set_completed_system
```

src/tweenable.rs
```diff
@@ -945,6 +1014,7 @@ pub struct Delay<T> {
     timer: Timer,
     on_completed: Option<Box<CompletedCallback<Delay<T>>>>,
     event_data: Option<u64>,
+    system_data: Option<SystemId>,
```
djeedai, line 1017:
```suggestion
    system_id: Option<SystemId>,
```

src/tweenable.rs
```diff
@@ -1041,6 +1112,44 @@ impl<T> Delay<T> {
...
+    /// fn test_system(query: Query<Entity, With<Enemy>>) {
+    ///    for enemy in query.iter() {
+    ///       println!("Found an enemy!");
+    ///    }
+    /// }
+    /// ```
+    ///
+    /// [`with_completed_system()`]: Tween::with_completed_system
+    #[must_use]
+    pub fn with_completed_system(mut self, user_data: SystemId) -> Self {
```
djeedai, line 1148:
```suggestion
    pub fn with_completed_system(mut self, system_id: SystemId) -> Self {
```

src/tweenable.rs
```diff
@@ -1100,6 +1209,28 @@ impl<T> Delay<T> {
...
+
+    /// Enable running a one-shot system upon completion.
+    ///
+    /// If enabled, the tween will run a system via a provided [`SystemId`] when the
+    /// animation completes. This is similar to the [`with_completed()`],
+    /// but uses a system registered by [`register_system()`] instead of a callback.
+    ///
+    /// See [`with_completed_system()`] for details.
+    ///
+    /// [`set_completed_system()`]: Tween::set_completed_system
```
djeedai, line 1221: You never need to tell rustdoc about the current function being documented. You do however need to help rustdoc with other functions:

```suggestion
    /// [`with_completed_system()`]: Tween::with_completed_system
```

src/tweenable.rs
```diff
@@ -1203,7 +1345,16 @@ mod tests {
...
-                tween.tick(duration, &mut target, entity, &mut events)
+                // let command_queue = &mut CommandQueue::default(); // todo
+                // let mut asd = Commands::new(command_queue, world);
+                tween.tick(
+                    duration,
+                    &mut target,
+                    entity,
+                    &mut events,
+                    // passing dummy values to let things compile
+                    &mut Commands::new(&mut CommandQueue::default(), &World::default()),
```
djeedai, line 1356: You can use (from Bevy's own code):

```rust
let mut command_queue = CommandQueue::default();
let mut commands = Commands::new(&mut command_queue, &world);
```

That way we could even inspect the command queue to confirm the sytem ran (or, apply it to the world and make sure the system ran by setting a flag when it does and checking the flag after the fact).

djeedai (changes requested):
I'm conceptually on board but there's a number of minor things to fix. Let me know if you don't have time to, I can take over and finish it. Thanks!

src/tweenable.rs
Peepo-Juice: Apologies, I clearly don't know what im doing lol.
Would "[`bevy::ecs::world::World::register_system`]" be acceptable? It seems Rustdoc understands it when I include the full path.

Peepo-Juice: Sorry, I just realized a better way to do that. Disregard my last comment.

src/tweenable.rs
Peepo-Juice: Have you tried this? Borrow checker not letting me do this.

djeedai (approved):

djeedai (approved):

PR #112, opened 2023-11-27
Add `SpriteAlphaLens`

johanhelsing:
Sometimes it's convenient not to overwrite the color of a sprite. If they have conflicting animations or are controlled by other systems for instance.

Makes it a lot simpler to create simple "fade out" animations:

```rust
                    .insert(Animator::new(Tween::new(
                        Linear,
                        Duration::from_secs_f32(0.3),
                        SpriteAlphaLens { start: 1., end: 0. },
                    )))
```


djeedai:
Hi, thanks for the contribution @johanhelsing. In general unfortunately I'd like to avoid supporting this kind of basic lens. There's plenty of other lenses that would be useful too, and the principle is that the crate allows users to easily define their own, rather than providing a fixed set of functionalities. The current ones are almost more like examples rather than actual core features of the crate. The problem with adding all those lenses is that this adds up to the cost of maintenance of the crate (documentation, testing, examples, _etc._), with limited benefit (it's a few lines that most people can write, as opposed to providing a feature which is costly or complex to develop). This is not a judgement on the quality or usefulness of your code, which makes absolutely sense for the use case you describe, but rather a reflection of the cost/benefit analysis of being a single maintainer and having limited time I'd rather dedicate to solving issues which actually save substantial time to users.

djeedai:
Closing this without merging. As mentioned this is very easy (and encouraged) as user code, without adding to the maintenance of this crate. Thanks!

PR #119, opened 2024-02-22
Update to bevy 0.13

dgsantana:
This updates dependecies to bevy 0.13.

Added a private trait to lerp colors, to fix the missing conversion between Color and Vec4.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,11 +24,11 @@ bevy_ui = ["bevy/bevy_ui", "bevy/bevy_render"]
 bevy_text = ["bevy/bevy_text", "bevy/bevy_render", "bevy/bevy_sprite"]
 
 [dependencies]
-interpolation = "0.2"
-bevy = { version = "0.12", default-features = false }
+interpolation = "0.3"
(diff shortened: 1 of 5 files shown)


djeedai (approved):

djeedai:
Check the features, you don't need all those dead code exceptions, it's just that `Color` is only used when some feature is which is why the compiler complains rightfully when the feature is turned off. CI tests all feature independently so will catch those errors.

dgsantana:
It's weird because the things it's complaining I didn't touch, that's what I find it weird.

djeedai:
> It's weird because the things it's complaining I didn't touch, that's what I find it weird. I can add cfg feature's to try and fix this.

Likely the linter has evolved since last time the code was built and merged, and so is picking up new lints which create new errors. I bet re-running CI on the main branch would fail with the same errors. Not your fault, and an unfortunate timing. I can pick that PR up later if you can't fix it or don't have time.

dgsantana:
If you can please look. I did run this on my Linux and Windows machine all the CI tests and didn't detect any problem.

djeedai (approved):

PR #127, opened 2024-06-19
Added lerping for all variants of Val

SenneDeVocht:
Val::Vw, Val::Vh, Val::VMin, and Val::VMax were missing from the lerp_val function.


djeedai (approved):
Thanks!

PR #133, opened 2024-07-27
Trigger observers when TweenCompleted event is sent

Peepo-Juice:
This PR is very simple: trigger observers when the TweenCompleted event gets sent out. I added this right next to event.send() so it will always happen if use "with_completed_event".

However, we may want to add an equivalent to "with_completed_event" so its opt-in, i.e. "with_completed_trigger" and "with_completed_trigger_targets".

For now, this is the easiest solution to get the ball rolling on this.

### DIFF
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -801,7 +801,7 @@ impl<T> Tweenable<T> for Tween<T> {
 
 
                 // trigger all entity-scoped observers
-                commands.trigger_targets(event, [entity]);
+                commands.trigger_targets(event, entity);
             }
             if let Some(cb) = &self.on_completed {
                 cb(entity, self);
--- a/src/tweenable.rs
+++ b/src/tweenable.rs
@@ -1336,11 +1336,8 @@ impl<T> Tweenable<T> for Delay<T> {
                 // send regular event
                 events.send(event);
 
-                // trigger all global observers
-                commands.trigger(event);
-
                 // trigger all entity-scoped observers
-                commands.trigger_targets(event, [entity]);
+                commands.trigger_targets(event, entity);
             }
             if let Some(cb) = &self.on_completed {
                 cb(entity, self);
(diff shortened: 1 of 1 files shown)


musjj:
@djeedai 

Any chance this can get merged soon? Observers make it a lot easier to deal with entity-based events. It looks like that the only failing check is the decreased coverage, but I don't see how that can be improved (false negative?).

src/tweenable.rs
```diff
@@ -791,10 +791,19 @@ impl<T> Tweenable<T> for Tween<T> {
...
                     entity,
                     user_data: *user_data,
-                });
+                };
+
+                // send regular event
+                events.send(event);
+
+                // trigger all global observers
+                commands.trigger(event);
```
musjj, line 803: This shouldn't be necessary, world-scoped observers should be able to see entity-scoped triggers too.

```suggestion
```

src/tweenable.rs
```diff
@@ -791,10 +791,19 @@ impl<T> Tweenable<T> for Tween<T> {
...
+
+                // trigger all entity-scoped observers
+                commands.trigger_targets(event, [entity]);
```
musjj, line 806: A small nit, but you don't need an array:
```suggestion
                commands.trigger_targets(event, entity);
```

src/tweenable.rs
```diff
@@ -1321,10 +1328,19 @@ impl<T> Tweenable<T> for Delay<T> {
...
```
musjj, line 1343: You forgot this one :slightly_smiling_face: 

```suggestion
                // trigger all entity-scoped observers
                commands.trigger_targets(event, entity);
```

src/tweenable.rs
```diff
@@ -791,10 +791,17 @@ impl<T> Tweenable<T> for Tween<T> {
...
+                let event = TweenCompleted {
                     entity,
                     user_data: *user_data,
-                });
+                };
+
+                // send regular event
+                events.send(event);
+
+
```
djeedai, line 802: Can you please remove that extra blank line, which will also re-trigger a CI against the latest version (with Bevy 0.15 support)? Thanks

djeedai (changes requested):

djeedai (approved):

Issue #115, opened 2024-01-22
Adding new tweens to existing ones


musjj:
Let's say I have an entity that is set up with a tween targeting the entity's `transform.rotation`. The tween loops infinitely throughout the game.

But I also want to have a system that can insert a new tween that targets said entity's `transform.scale`, depending on the state of the game. The tween loops only once. I also want to be able to replay that tween, depending on the game state.

The problem is that both of these tweens requires an `Animator<Transform>` component on the entity, which is currently not possible in Bevy. Is there a good solution for this use-case?

SecretPocketCat:
You could use tracks, but you'd have to recreate the looping tween which would be tedious.
A simple-ish approach would be to add a wrapper that can have it's own animator which can do the rotation.

guyguy2001:
A similar need I have (I'm not sure if this should go here or in a new issue) is to tell a tween to run after the current tween.
So for example, when my player is hit, I want them to flash red, and afterwards start animating their alpha to indicate an invulnerability state.
(This might not be the correct way to approach the problem, I'm not sure yet).

To do this I need one of the following:
* A `.then(self, &tween)` method on the `Tweenable` trait, and a way of owning `animator.tweenable()` in order to be able to call it (since `Sequence::then` only accepts an owned `self`, not a reference)
* A `.then` method on the animator itself (which will probably use the `.then` I suggested)
* A way of passing `animator.tweenable()` into `Sequence::from_single` (or `::new` or the likes).

I also need a way to know when the current animation will end - but I think I can get that via the existing `total_duration`, which seems great

The second option is probably the best one, but I'm too new to writing APIs in rust to have a clear idea.
Also, I could see a world in which it is just a bad idea to allow enqueuing more animations from the outside.

djeedai:
@musjj sorry I completely missed this issue.

> The problem is that this requires you to put more than one Animator<Transform> component on the entity, which is currently not possible in Bevy. Is there a good solution for this use-case?

[...]

PR #136, opened 2024-11-24
Allow `Animator` to target other entities

patrickariel:
This adds an optional property to `Animator` that stores an entity.
If the property contains a value, it will be used as the target of the animation. Otherwise, the `Animator`'s entity will be used (default behavior).

This makes the creation of complex, modular tweens much easier. For example, you can now have one animator entity tweening `Transform::translation` and the other tweening `Transform::rotation`. Previously, this would have to be done with a single animator, which is very inflexible.

They can be used like this:

```rust
let sprite = commands.spawn(
    SpriteBundle {
        sprite: Sprite {
            color: Color::RED,
            custom_size: Some(Vec2::new(size, size)),
            ..default()
        },
        ..default()
    },
).id();

// This animator will animate the previously spawned entity
commands.spawn(Animator::new(tween).with_target(sprite));
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -537,7 +537,7 @@ impl<T: Component> Animator<T> {
         }
     }
 
-    /// Create a new version of this animator with the [target] set to the given entity.
+    /// Create a new version of this animator with the `target` set to the given entity.
     pub fn with_target(mut self, entity: Entity) -> Self {
         self.target = Some(entity);
         self
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -511,7 +511,8 @@ macro_rules! animator_impl {
 pub struct Animator<T: Component> {
     /// Control if this animation is played or not.
     pub state: AnimatorState,
-    target: Option<Entity>,
+    /// When set, the animated component will be the one located on this entity.
+    pub target: Option<Entity>,
     tweenable: BoxedTweenable<T>,
     speed: f32,
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -510,6 +510,7 @@ macro_rules! animator_impl {
 pub struct Animator<T: Component> {
     /// Control if this animation is played or not.
     pub state: AnimatorState,
+    target: Option<Entity>,
     tweenable: BoxedTweenable<T>,
     speed: f32,
 }
@@ -529,8 +530,15 @@ impl<T: Component> Animator<T> {
         Self {
             state: default(),
             tweenable: Box::new(tween),
+            target: None,
             speed: 1.,
         }
+    }
+
+    /// Set a custom target entity for the animator.
+    pub fn with_target(mut self, entity: Entity) -> Self {
+        self.target = Some(entity);
+        self
     }
 
     animator_impl!();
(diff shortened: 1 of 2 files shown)


djeedai:
> This makes the creation of complex, modular tweens much easier. For example, you can now have one animator entity tweening Transform::translation and the other tweening Transform::rotation. Previously, this would have to be done with a single animator, which is very inflexible.

I don't understand this rationale. Either you want to animate both translation and rotation at the same time, and then you use a combined `Lens`. Or you want to have completely unrelated translation and rotation, for example your object moves along some kind of trajectory, and also spins on itself, and chances are that's because you conceptually have 2 different objects and should have 2 different entities, with one parented to the other (the spinning child entity parented to the trajectory entity).

Can you please provide some detailed example where that feature is useful? I'm all for adding new features, but this makes the code more complex, adds some indirection query, for a result which it seems can be achieved without any change.

patrickariel:
Yes, that particular example can be worked around by re-parenting, but my general rationale is that you sometimes want to target different fields of the same component via different animators.

For example, let's say I want to create a button with a pulsing border width that loops indefinitely (as a way to attract player attention). I would need to insert an `Animator<Node>` that targets [`Node::border`](https://docs.rs/bevy/latest/bevy/prelude/struct.Node.html#structfield.border). Now when the user hovers over the button, I want it to bounce up a bit. This requires inserting another temporary `Animator<Node>` that targets [`Node::bottom`](https://docs.rs/bevy/latest/bevy/prelude/struct.Node.html#structfield.bottom). But of course, this isn't possible in ECS.

Maybe in a perfect world, components will always be split up in such a granular way that a problem like this will never happen. But in messy, complex scenarios, something like this will *eventually* be a problem. Especially when dealing with external libraries that you can't control. And when the time comes, you want your users to have the power to deal with it.

As a precedent, [`bevy_eventlistener`](https://github.com/aevyrie/bevy_eventlistener) had a similar [issue](https://github.com/aevyrie/bevy_eventlistener/issues/21). The problem was that the event handler entity and the event target entity was *coupled*. This means that you can't put more than one handler component per event type in the target entity.

But when it was upstreamed as observers, this problem was solved by decoupling the *observer* entity and the *observed* entity. So now an entity can have any number of observers of the same event type.
This library is currently also having the exact same [issue](https://github.com/djeedai/bevy_tweening/issues/115): the *animator* entity and the *animated* entity is coupled together, creating problems like mentioned above.

Another limitation I've found is that you can't animate more than one component per animator. Which means that it's hard to make tween sequences that alternates between different components. Of course, something like this will require a massive rework, so it's pretty unlikely that it can be done in this library.

As a comparison, tweening APIs from other game engines don't have any of the above limitations: see [Godot](https://docs.godotengine.org/en/stable/classes/class_tween.html), [Unity](https://docs.unity3d.com/Packages/com.unity.tiny@0.16/api/Unity.Tiny.Tweens.TweenSystem.html), etc.

For the reasons above I've since moved to [`bevy_tween`](https://github.com/Multirious/bevy_tween), which solved most of my gripes. So if you think this PR is out of place, I'm completely fine with it being closed :slightly_smiling_face:

djeedai (approved):

PR #138, opened 2024-12-02
Bevy 0.15 support

rendaoer:
The main content is to support Bevy 0.15, switch from Style to Node, use TextColor for Text, etc., Migrating from the interpolation library to Bevy, modify the Bundle part of the example, check and optimize the code and documents


src/plugin.rs
```diff
@@ -29,7 +31,7 @@ use crate::{tweenable::ComponentTarget, Animator, AnimatorState, TweenCompleted}
 ///
 /// [`Transform`]: https://docs.rs/bevy/0.12.0/bevy/transform/components/struct.Transform.html
 /// [`Text`]: https://docs.rs/bevy/0.12.0/bevy/text/struct.Text.html
-/// [`Style`]: https://docs.rs/bevy/0.12.0/bevy/ui/struct.Style.html
+/// [`Node`]: https://docs.rs/bevy/0.15.0/bevy/ui/struct.Node.html
 /// [`Sprite`]: https://docs.rs/bevy/0.12.0/bevy/sprite/struct.Sprite.html
 /// [`ColorMaterial`]: https://docs.rs/bevy/0.12.0/bevy/sprite/struct.ColorMaterial.html
```
mnmaita, line 36: Maybe these links should point to `latest` docs? And also I think `Text` should be swapped for `TextColor` here too.

README.md
```diff
@@ -60,8 +60,8 @@ By default, 🍃 Bevy Tweening adopts a minimalist approach, and the `TweeningPl
 | `Transform`          | Yes                           |
 | `Sprite`             | Only if `bevy_sprite` feature |
 | `ColorMaterial`      | Only if `bevy_sprite` feature |
-| `Style`              | Only if `bevy_ui` feature     |
-| `Text`               | Only if `bevy_text` feature   |
+| `Node`              | Only if `bevy_ui` feature     |
+| `TextColor`               | Only if `bevy_text` feature   |
```
mnmaita, line 64: Minor nit, but check if you can align the table here.

README.md
```diff
@@ -168,7 +165,7 @@ Asset animation always requires the `bevy_asset` feature.
```
mnmaita, line 165: I can't comment above this but in L155 `Style` should be replaced with `Node` and in L157 `Text` with `TextColor` probably?

examples/menu.rs
```diff
@@ -43,32 +43,29 @@ fn main() {
...
+                padding: UiRect::all(Val::Px(16.)),
+                flex_direction: FlexDirection::Column,
+                align_content: AlignContent::Center,
+                align_items: AlignItems::Center,
+                align_self: AlignSelf::Center,
+                justify_content: JustifyContent::Center,
                 ..default()
             },
-            Name::new("menu"),
+            BackgroundColor(Color::NONE),
```
mnmaita, line 68: I think this is not really necessary anymore as `Node`s no longer have a `BackgroundColor` by default, but please correct me if I'm wrong.

examples/menu.rs
```diff
@@ -98,39 +95,34 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
-                                align_content: AlignContent::Center,
-                                align_items: AlignItems::Center,
-                                align_self: AlignSelf::Center,
-                                justify_content: JustifyContent::Center,
-                                ..default()
-                            },
-                            background_color: BackgroundColor(NORMAL_COLOR),
-                            transform: Transform::from_scale(Vec3::splat(0.01)),
+                        Name::new(format!("button:{}", text)),
+                        Node {
```
mnmaita, line 99: This is missing the `Button` component AFAICS.

examples/sequence.rs
```diff
@@ -187,22 +183,20 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
         BlueSprite,
         Animator::new(seq2),
     ));
 }
 
 fn update_text(
-    mut query_text_red: Query<&mut Text, (With<RedProgress>, Without<BlueProgress>)>,
-    mut query_text_blue: Query<&mut Text, (With<BlueProgress>, Without<RedProgress>)>,
+    query_text_red: Query<&Children, (With<RedProgress>, Without<BlueProgress>)>,
+    query_text_blue: Query<&Children, (With<BlueProgress>, Without<RedProgress>)>,
```
mnmaita, line 198: You could be using `Single` on these two instead of `Query`, and I think the mutually exclusive filters are not needed anymore due to the change of mutability of the query. Just a minor and optional suggestion.

src/tweenable.rs
```diff
@@ -446,7 +446,7 @@ impl<T: 'static> Tween<T> {
     /// # Example
     /// ```
     /// # use bevy_tweening::{lens::*, *};
-    /// # use bevy::math::*;
+    /// # use bevy::math::{*,curve::EaseFunction};
```
mnmaita, line 449: Could you please format all these doc changes (i.e. adding a space after each comma, etc.)?

rendaoer: OK

rendaoer:
@mnmaita Thank you for your code review

djeedai (commented):
Thanks for the comprehensive PR. I'll try to review in details when I have more time, likely later this week or on the week-end, but it seems it's chewing a lot. There's the upgrade, and then there's unrelated things like adding a prelude (I think?), changing the easing functions, _etc._ In general I'd prefer to get standalone PRs for each bit, that not only makes it easier to review, but also makes it easier to bisect bugs in the future and potentially revert changes.

rendaoer:
OK, I'll split them later.

rendaoer:
But isn't EaseFunction considered included in the range of 0.15?

rendaoer:
In places like this,
```rust
use bevy::prelude::*;
use bevy_tweening::{lens::*, *};
```
bevy::prelude::* exports bevy's EaseFunction, and then bevy_tweening::{lens::*, *} also exports interpolation's EaseFunction, and this happens in many places, and in the documentation, making it very cumbersome to modify. And I thought this was part of the migration to 0.15?

djeedai:
Need to add `x11` to Bevy features otherwise Linux won't build (`winit` needs either `x11` or `wayland`; see https://github.com/rust-windowing/winit/issues/3174).

src/plugin.rs
djeedai: No they shouldn't. If they point to `latest`, then when you look at the docs for version N you get a link to the latest version, which is only N until N+1 is published, and will after that forever be wrong.

djeedai:
Thanks @rendaoer and @mnmaita for this. I merged the change now, I want to move ahead and release.

musjj:
@rendaoer 
Are you planning to open another PR for the prelude module?

djeedai:
> @rendaoer Are you planning to open another PR for the prelude module?

This PR removed the dependency to `interpolation` so there's no issue with the prelude, is there?

musjj:
Yeah, but it's more idiomatic to have a prelude (`use bevy_tweening::{lens::*, *}` vs. `use bevy_tweening::prelude::*`. Your other crate `bevy_hanabi` also uses a prelude, so I don't see a reason not to have one here too.

djeedai:
@musjj ah sorry I thought you were talking about the symbol conflict on `EaseFunction`. I don't mind having a prelude, although the crate is fairly small so this is not really needed I think.

PR #141, opened 2024-12-20
Move enabling of Bevy X11 into separate default feature

### LINKED ISSUE #140: x11 feature
is it possible to allow the user of `bevy_tweening` to optionally enable `x11` only if needed?

TrialDragon:
Resolves #140

This way those who want to avoid bigger binary sizes in Wasm can disable a feature and reduce the size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,9 @@ readme = "README.md"
 exclude = ["examples/*.gif", ".github", "release.md", "run_examples.bat"]
 
 [features]
-default = ["bevy_sprite", "bevy_ui", "bevy_asset", "bevy_text"]
+default = ["bevy_sprite", "bevy_ui", "bevy_asset", "bevy_text", "x11"]
+# Enable the X11 display server for Linux builds
+x11 = ["bevy/x11"]
 # Enable support for Asset animation
 bevy_asset = ["bevy/bevy_asset"]
 # Enable built-in lenses for Bevy sprites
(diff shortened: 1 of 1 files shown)


djeedai:
Did you actually check that this makes a difference? Because I was looking at whether we can make that feature only for Linux and found that post specifically about `winit` saying that `x11` doesn't do anything on other platforms.

https://stackoverflow.com/questions/67373380/can-i-set-cargo-projects-default-features-depending-on-the-platform#75780957

TrialDragon:
Hmm, on second look it seems it does indeed do nothing on any non unix platform per their build.rs which defines the `x11_platform` cfg alias such that the x11 things are only compiled if all of the listed happen: A.) the x11 feature is enabled, B.) the target is of the unix kind and not apple or emscripten, and C. if it isn't redox.

https://github.com/rust-windowing/winit/blob/f0d8689039b994ea404ae4d237dfd25af0b5672b/build.rs#L19

There is two raw usages of the x11 feature in the XKB module (https://github.com/rust-windowing/winit/blob/f0d8689039b994ea404ae4d237dfd25af0b5672b/src/platform_impl/linux/common/xkb/mod.rs#L38), but that's not really worth it imo.

Put simply you're right and this should barely effect non-unix builds.

djeedai:
Great, thanks for confirming, it's always good to make sure of those things and keep wasm slim! 😊

PR #148, opened 2025-05-30
Fix panic when constructing empty tracks

jim-ec:
Fall back to a duration of zero if no tweenables are given to the tracks constructor.


djeedai:
Closing as Obsolete, `Tracks` has been removed.

PR #149, opened 2025-05-30
Implement Identity Tween

jim-ec:
A simple no-op tween that I found useful during development, similar to `Transform::IDENTITY`.


djeedai:
Closing this, this is outdated and doesn't work with the latest generic-less design.

In general I'm not sure why you'd want an identity tween since there's already several ways to "disable" a tween. At the very worst (IMHO) spawn it with a `Delay(0)` instead. But really I'd like to understand a pattern where you need to spawn right now but don't know about the lens yet, I can't see why you'd be in that situation.

PR #162, opened 2025-12-09
UITransform translation lens

extrawurst:
opening this as a RFC:

as it only allows `px` right now.

### DIFF
--- a/src/lens.rs
+++ b/src/lens.rs
@@ -452,6 +452,27 @@ impl Lens<UiTransform> for UiTransformRotationLens {
     }
 }
 
+/// A lens to manipulate the [`translation`] field of a [`UiTransform`] component.
+///
+/// [`translation`]: https://docs.rs/bevy/0.17/bevy/ui/ui_transform/struct.UiTransform.html#structfield.translation
+/// [`UiTransform`]: https://docs.rs/bevy/0.17/bevy/ui/ui_transform/struct.UiTransform.html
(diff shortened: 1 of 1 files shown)


src/lens.rs
```diff
@@ -448,7 +448,28 @@ pub struct UiTransformRotationLens {
 #[cfg(feature = "bevy_ui")]
 impl Lens<UiTransform> for UiTransformRotationLens {
     fn lerp(&mut self, mut target: Mut<UiTransform>, ratio: f32) {
-        target.rotation = self.start.slerp(self.end, ratio);
+        target.rotation = self.start.nlerp(self.end, ratio);
```
djeedai: There's no reason to use `nlerp()` in my opinion, this will distort the curve. `lerp()` should return a linear interpolation, which for a rotation means linear angular motion.

djeedai (changes requested):
Thanks for the PR! I don't mind adding that lens, seems animating UI is a very common use case, so I'm fine paying the maintenance cost for that one. The "limited to px" also, it's too complex to handle all UI values and generally undefined anyway. However there's a few issues:
- branch has conflicts (my bad)
- unless there's an overwhelming concern (with profiling data) about `nlerp()` I much rather have something "correct" like `slerp()` instead

extrawurst:
@djeedai ok conflicts and comments resolved I think.

djeedai (approved):

Issue #169, opened 2026-06-22
bevy 0.19 support


extrawurst:
would be great to update

PR #170, opened 2026-06-22
Update to bevy 0.19

spebern:
Bumps the `bevy` dependency from 0.18 to 0.19 and fixes the resulting breakage.

### Library
- `EntityCommand` gained an associated `type Out`; added `type Out = ();` to the 10 transform command impls.
- `Components::{resource_id, get_resource_id}` renamed to `{component_id, get_id}`.
- `Assets::get_mut` now returns `AssetMut`; the internal asset resolver uses `get_mut_untracked` (it already does its own change detection via `filter_map_unchanged`).

### Tests
- `SystemState::get_mut` now returns a `Result` (`.unwrap()`).
- Asset-lens test helpers use `Assets::get_mut_untracked`.

### Examples
- `TextFont::{font, font_size}` are now `FontSource`/`FontSize` (`.into()`).
- `TextLayout::new_with_justify` removed; replaced with a struct literal.
- `AmbientLight` is now a per-camera component and `GlobalAmbientLight` is the global resource — the `ambient_light` example animates `GlobalAmbientLight`, preserving the resource-animation intent.
- `Hdr` moved to `bevy::camera`.

Also bumps the `bevy-inspector-egui` dev-dependency 0.36 → 0.37.

All 63 tests pass and every example compiles against bevy 0.19.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -100,7 +100,7 @@ jobs:
           components: rustfmt, clippy
       - name: Install cargo-tarpaulin
         run: |
-          RUST_BACKTRACE=1 cargo install --version 0.31.2 cargo-tarpaulin
+          RUST_BACKTRACE=1 cargo install --version 0.32.8 cargo-tarpaulin
       - name: Generate code coverage
         run: |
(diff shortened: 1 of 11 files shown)


extrawurst:
closes #169

djeedai (approved):
I fixed the asset change detection to only trigger if the asset is actually modified by the animation (that is, by `Lens::lerp()` itself).

PR #171, opened 2026-06-25
Updated to Bevy 0.19.

Cyannide:
The fix that needs a review is the change detection for Assets especially to make colormaterial_color functional.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -283,10 +283,10 @@
 //! shine for simpler animations while `bevy_animation` while offer a more
 //! extensive support for larger, more complex ones.
 //!
-//! [`Transform::translation`]: https://docs.rs/bevy/0.18/bevy/transform/components/struct.Transform.html#structfield.translation
-//! [`Entity`]: https://docs.rs/bevy/0.18/bevy/ecs/entity/struct.Entity.html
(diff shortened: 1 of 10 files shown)


Cyannide:
#170 the colormaterial_color example will not work because there is no AssetEvent::Modified triggered to push the material update to the GPU.

Cyannide:
Also fixes #169

src/lib.rs
```diff
@@ -2681,10 +2701,10 @@ impl TweenResolver {
             let asset_id = asset_id.typed::<A>();
             // First, remove the Assets<A> from the world so we can access it mutably in
             // parallel of the TweenAnim
-            world.resource_scope(|world, assets: Mut<Assets<A>>| {
-                // Next, fetch the asset A itself from its Assets<A> based on its asset ID
-                let Some(asset) = assets.filter_map_unchanged(|assets| assets.get_mut(asset_id))
-                else {
+            world.resource_scope::<Assets<A>, _>(|world, mut assets: Mut<Assets<A>>| {
+                // get_mut with into_inner() triggers AssetEvent::Modified
+                // essential for the render pipeline to re-upload to GPU
```
djeedai, line 2706: Essential but completely breaking the design which carefully avoids marking assets as changed all the way through to `Lens::lerp()`. This change unconditionally marks the asset modified, whether or not it's actually animated. See my fix on #170.

PR #172, opened 2026-06-28
Other Bevy 0.19 updates that were part of the #171 PR...

Cyannide:
Here are the other updates that I had made for the PR. Including doc updates to some moved links etc.

I also think that when specifying font size in the examples, you should now declare the kind of sizing you're using...ie, px em etc since it is available. Not just use .into() for the default because the default could change.

Thanks for the great plugin!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,6 @@ required-features = [
     "bevy/hdr",
     "bevy/tonemapping_luts",
     "bevy/bevy_post_process",
-	"bevy/zstd_rust",
 ]
 
 [[example]]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -54,6 +54,7 @@ required-features = [
     "bevy/hdr",
     "bevy/tonemapping_luts",
     "bevy/bevy_post_process",
+	"bevy/zstd_rust",
```
djeedai: I don't understand this change. It doesn't seem necessary.

Cyannide: When I was originally working on the update, Bevy wanted the zstd backend selected. Doesn't seem to require it now though so I removed it.

djeedai:
Thanks for the PR, but there are a few changes I only do myself at the time of release, like bumping the crate version or README, otherwise they confuse users. I've released 0.16 now. Closing this as obsolete.

